PR #26, opened 2024-01-31
Add a container to allow wp-cli over ssh

EarthlingDavey:
This will allow the migration script (in a different local project) to use wp-cli remotely.

It can be used with the following arguments: `wp --ssh=ssh-user@host.docker.internal:2222/var/www/html db check` and will be useful for `import`ing from an xml file.

It cannot use dory because it's not http traffic. So it exposes the port to the host, and ssh commands for other containers use host.docker.internal

@wilson1000-MoJ do you mind taking a look, please.
- Is it ok to use a hardcoded password if this is only intended to be run locally?
- What's your opinion on me creating a dedicated container for this?
- Could you suggest a good way to add this to the Makefile. Currently I'm doing `docker compose --profile ssh up`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,22 @@ COPY deploy/config/local/pool.conf /usr/local/etc/php-fpm.d/pool.conf
 # www-data
 USER 82
 
+
+## target: ssh
+FROM base AS ssh
+
+RUN apk add --no-cache openssh bash
+
+RUN ssh-keygen -A 
+RUN adduser -h /home/ssh-user -s /bin/bash -D ssh-user
+RUN echo 'PasswordAuthentication yes' >> /etc/ssh/sshd_config
+RUN echo -n "ssh-user:test" | chpasswd
+RUN echo 'cd /var/www/html' >> /home/ssh-user/.bash_profile
+
+EXPOSE 22
+
+CMD ["/usr/sbin/sshd", "-D", "-e"]
+
 ###
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,14 +23,17 @@ USER 82
 
 
 ## target: ssh
-FROM base AS ssh
+FROM base AS local-ssh
+
+ARG LOCAL_SSH_PASSWORD
 
 RUN apk add --no-cache openssh bash
 
 RUN ssh-keygen -A 
 RUN adduser -h /home/ssh-user -s /bin/bash -D ssh-user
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -21,6 +21,22 @@ COPY deploy/config/local/pool.conf /usr/local/etc/php-fpm.d/pool.conf
...
+
+## target: ssh
+FROM base AS ssh
+
+RUN apk add --no-cache openssh bash
+
+RUN ssh-keygen -A 
+RUN adduser -h /home/ssh-user -s /bin/bash -D ssh-user
+RUN echo 'PasswordAuthentication yes' >> /etc/ssh/sshd_config
+RUN echo -n "ssh-user:test" | chpasswd
```
wilson1000-MoJ, line 33: Just to air on the side of caution, our cyber-security team scan our repos. Can we include the password in the .env file and reference it here with something like the following?

.env
```
MIGRATE_SSH_PASSWORD=test
```

Dockerfile
```
ARG MIGRATE_SSH_PASSWORD

...
RUN echo -n "ssh-user:${MIGRATE_SSH_PASSWORD}" | chpasswd
...
```

Makefile
```
migrate:
	docker compose up --profile ssh
```

wilson1000-MoJ: I'm not sure if that will work due to the funny way Compose and Dockerfile manage environments, but worth a go. The service should run so try:

`docker compose exec -it ssh -- bash`

wilson1000-MoJ: ... and maybe check that `chpasswd` did what you wanted 👍🏽

EarthlingDavey: Thanks @wilson1000-MoJ ! I didn't know the security team would scan the repos. That's good to know.

I've made the changes, except I used `LOCAL_SSH_PASSWORD` (before I saw your comment). Is that ok, as I feel `LOCAL` is a good indication to only use it in a local environment.

Happy to rename to `MIGRATE_SSH_PASSWORD` if you prefer.

wilson1000-MoJ: `LOCAL_SSH_PASSWORD` Sounds good to me :)

EarthlingDavey: Nice, it's done & working nicely here, do you mind doing a final review?

wilson1000-MoJ (commented):
Looks good :smile:

wilson1000-MoJ (approved):

PR #27, opened 2024-01-31
CDPT-1314 Add Gov Notify service

wilson1000-MoJ:
**_Adds two scripts_**
1. mail.php - includes the `wp_mail` override functions
2. mail-templates.php - contains all available templates, prepared on Gov Notify

Updates `deploy.yml` and all associated `secret.tpl` files to inject an API secret.

### DIFF
--- a/public/app/themes/justice/inc/mail-templates.php
+++ b/public/app/themes/justice/inc/mail-templates.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * G O V . U K   N O T I F Y
+ * -.-.-.-.-.-.-.-.-.-.-.-.-
+ * Template configuration for WordPress system
+ * Templates can be found on the GovUK Notify website:
+ * https://www.notifications.service.gov.uk/
+ */
+return [
+
+    /**
+     * Email configuration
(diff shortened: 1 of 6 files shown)


public/app/themes/justice/inc/mail-templates.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+return [
+
+    /**
+     * Email configuration
+     ************************/
+
+    'email' => [
+        'default' => [
+            'id' => '0e73e29f-2efa-4313-a6e5-8908af51a6cc',
```
EarthlingDavey, line 17: I haven't worked with notifications before. Will this ID be consistent across all environments? If it's not, we should consdiser using an environment variable.

EarthlingDavey (approved):
It looks good, just the one comment about env. vars.

public/app/themes/justice/inc/mail-templates.php
wilson1000-MoJ: Hey 😄 good question. The email templates are persistent across environments. We're good to list templates like this.

PR #31, opened 2024-02-06
Add WordPress importer

wilson1000-MoJ:


EarthlingDavey:
What do you think about adding something like the following code to disable the importer plugin on anything but development...

```php
<?php 
/*
 * Plugin Name: Disable/Configure Plugins
 * Description: Lets you disable/configure plugins based on environment variables
 * Author:      Chris Sewell
 * @see https://wordpress.stackexchange.com/a/281114
 * 
 * This is a "Must-Use" plugin. Code here is loaded automatically before regular plugins load.
 * This is the only place from which regular plugins can be disabled programmatically.
 */

/* Disable specified plugins in non-development environments */
if (getenv('WP_ENV') !== 'development' && is_admin()) {
    $plugins = array(
        'wordpress-importer/wordpress-importer.php'
    );
    require_once(ABSPATH . 'wp-admin/includes/plugin.php');
    deactivate_plugins($plugins);
}
```

And, not specifically related to this commit, but to the work in https://github.com/ministryofjustice/justice-gov-uk/pull/30 .

To get nginx working locally we need to add to the docker file

```dockerfile
FROM base-nginx AS nginx-dev
# Add the following line or similar...
RUN echo "# This is a placeholder, because the file is `include`d in `php-fpm.conf`." > /etc/nginx/server_name.conf
```

EarthlingDavey (approved):

PR #35, opened 2024-02-07
Feat to the top block

EarthlingDavey:
@wilson1000-MoJ Do you mind taking a look? I'm open to discussion on any of the choices I've made.

- Create a blocks folder in src. This will be useful for when it makes sense to keep .js and .css together.
- Created an `admin.js` file, enqueued that along with the frontend `app.css` file.
- The Back to top button is literally the most simple block I could make. There are no custom attributes, like size/colour etc.

It means I can add the following markup script to the migrate script. So as to combine the various ways that the old website implemented the feature into one. Easier for devs to manage & easy for admins (type `/ top` in the page editor).

```html
<!-- wp:moj/to-the-top -->
<a href="top" class="to-the-top">To the top</a>
<!-- /wp:moj/to-the-top -->
```


wilson1000-MoJ (commented):
LGTM 👍🏽 😄

wilson1000-MoJ (approved):

PR #39, opened 2024-02-09
WP Offload Media - settings

EarthlingDavey:
Add Minio for local development, it helps not using AWS locally for debugging.

Add verify, and document S3 & CloudFront settings.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -147,3 +147,17 @@ COPY --from=assets-build /code/public/app/themes/justice/dist /var/www/html/publ
 COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html/vendor-assets /var/www/html/
 COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html/public/index.php /var/www/html/public/index.php
 COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html/public/wp/wp-admin/index.php /var/www/html/public/wp/wp-admin/index.php
+
+
+###
+
+## target: local-cdn
+FROM caddy:2.7.6-alpine AS local-cdn
+
+# A CDN to proxy requests to the Minio (S3) server.
+# Mimics AWS CloudFront, and remove the bucket path from the URL.
+# e.g. Request: http://cdn.justice.docker/uploads/2024/02/xyz.jpg
+# proxies to:   http://minio:9000/bucket-name/uploads/2024/02/xyz.jpg
+RUN  echo $':2019 \n\
+rewrite * /{$S3_BUCKET_NAME}{uri} \n\
+reverse_proxy minio:9000' >  /etc/caddy/Caddyfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -155,9 +155,9 @@ COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html/public/wp
 FROM caddy:2.7.6-alpine AS local-cdn
 
 # A CDN to proxy requests to the Minio (S3) server.
-# Mimics AWS CloudFront, and remove the bucket path from the URL.
+# Mimics AWS CloudFront, and removes the bucket path from the URL.
 # e.g. Request: http://cdn.justice.docker/uploads/2024/02/xyz.jpg
-# proxies to:   http://minio:9000/bucket-name/uploads/2024/02/xyz.jpg
-RUN  echo $':2019 \n\
-rewrite * /{$S3_BUCKET_NAME}{uri} \n\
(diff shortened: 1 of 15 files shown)


wilson1000-MoJ (approved):

config/wp-offload-media.php
```diff
@@ -0,0 +1,66 @@
...
+
+// Merge in the access key and secret if they are set
+if (env('AWS_ACCESS_KEY_ID') && env('AWS_SECRET_ACCESS_KEY')) {
+    $as3_settings = array_merge($as3_settings, [
+         // Access Key ID for Storage Provider (aws and do only, replace '*')
+        'access-key-id' => env('AWS_ACCESS_KEY_ID'),
+        // Secret Access Key for Storage Providers (aws and do only, replace '*')
+        'secret-access-key' => env('AWS_SECRET_ACCESS_KEY')
+    ]);
+} else {
```
wilson1000-MoJ, line 59: I've got a challenge for you 😄 ... try and prevent else statements.
`switch` is faster than if/else however, if you have a case of this `var` is this, else, it is this. You can set the var and move to change it if the evaluation is true.

```
// default, production
$as3_settings['use-server-roles'] = true;

// Merge in the access key and secret if they are set
if (env('AWS_ACCESS_KEY_ID') && env('AWS_SECRET_ACCESS_KEY')) {
    $as3_settings['use-server-roles'] = false;
    $as3_settings = array_merge($as3_settings, [
        'access-key-id' => env('AWS_ACCESS_KEY_ID'),
        'secret-access-key' => env('AWS_SECRET_ACCESS_KEY')
    ]);
}
```

public/app/themes/justice/inc/amazon-s3-and-cloudfront-tweaks-for-minio.php
```diff
@@ -0,0 +1,172 @@
...
+
+        /*
+         * URL Rewrite related filters.
+         */
+        add_filter('as3cf_get_attachment_url', array( $this, 'getAttachmentUrl' ), 10, 4);
+        add_filter('as3cf_use_ssl', '__return_false', 10, 1);
+    }
+
+    // define some strings
+    private static $minio_host_in_backend_context = 'host.docker.internal';
```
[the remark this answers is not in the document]
EarthlingDavey: They should be minio & minio.justice.docker ... Good catch, I've updated them.

config/wp-offload-media.php
EarthlingDavey: Thanks, this is a really good point.

Dockerfile
```diff
@@ -147,3 +147,17 @@ COPY --from=assets-build /code/public/app/themes/justice/dist /var/www/html/publ
...
+## target: local-cdn
+FROM caddy:2.7.6-alpine AS local-cdn
+
+# A CDN to proxy requests to the Minio (S3) server.
+# Mimics AWS CloudFront, and remove the bucket path from the URL.
+# e.g. Request: http://cdn.justice.docker/uploads/2024/02/xyz.jpg
+# proxies to:   http://minio:9000/bucket-name/uploads/2024/02/xyz.jpg
+RUN  echo $':2019 \n\
+rewrite * /{$S3_BUCKET_NAME}{uri} \n\
+reverse_proxy minio:9000' >  /etc/caddy/Caddyfile
```
wilson1000-MoJ, line 163: I'm not familiar with the syntax here @EarthlingDavey - can you help explain? Sorry.

Would something like this work?:

```
RUN { \
       echo "$:2019" \
       echo "rewrite * /{$S3_BUCKET_NAME}{uri}" \
       echo "reverse_proxy minio:9000" \
    } > /etc/caddy/Caddyfile
```

wilson1000-MoJ (approved):

Dockerfile
wilson1000-MoJ: Feel free to resolve the convo to merge :) I'm just being curious here.
Thanks @EarthlingDavey - I'm super keen to see the effect on dev.

EarthlingDavey: Thanks for the suggestion, curly braces were erroring, I've changed it to `echo -e`. 

It's a new one to me, the -e flag enables interpretation of backslash escapes.

wilson1000-MoJ: Aw, the code was untested. Did you note the error, out of curiosity?

EarthlingDavey: > /bin/sh: syntax error: unexpected end of file (expecting "}")

Could the {}s b a bash thing? This container

wilson1000-MoJ (approved):

PR #45, opened 2024-02-15
Increase php limits to allow content import.

EarthlingDavey:
Increases file size for uploads and post requests. 
Increases execution time to prevent timeout during uploads.

### DIFF
--- a/deploy/config/local/php-fpm.conf
+++ b/deploy/config/local/php-fpm.conf
@@ -7,3 +7,5 @@ include server_name.conf;
 include fastcgi_params;
 fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
 fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+# Increase limits to allow for large file uploads. i.e. content import via Importer Plugin.
+fastcgi_param PHP_VALUE "upload_max_filesize=32M \n post_max_size=32M \n max_execution_time=300";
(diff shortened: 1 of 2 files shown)


deploy/config/local/php-fpm.conf
```diff
@@ -7,3 +7,5 @@ include server_name.conf;
 include fastcgi_params;
 fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
 fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+# Increase limits to allow for large file uploads. i.e. content import via Importer Plugin.
+fastcgi_param PHP_VALUE "upload_max_filesize=32M \n post_max_size=32M \n max_execution_time=300";
```
wilson1000-MoJ, line 11: DId you mean to enter \n in the file?

wilson1000-MoJ: Do you think it might be best to put this in a `user.ini` file?

```
upload_max_filesize=32M 
post_max_size=32M 
max_execution_time=300;
```

wilson1000-MoJ: ... or maybe something similar. Your call 😄

wilson1000-MoJ (approved):

EarthlingDavey:
For future reference ... we will revisit moving this into a dedicated file.

PR #52, opened 2024-02-19
Debug, remove CloudFront url.

EarthlingDavey:


deploy/development/deployment.tpl
```diff
@@ -45,11 +45,6 @@ spec:
               secretKeyRef:
                 name: s3-bucket-output
                 key: bucket_name
-          - name: CLOUDFRONT_URL
-            valueFrom:
-              secretKeyRef:
-                name: cloudfront-output
-                key: cloudfront_url
```
wilson1000-MoJ, line 52: This won't work. If you remove this, the deployment won't know what the URL is... and neither will we. This is currently the only method for us, and the application, to know safely, what the URL is.

EarthlingDavey:
Closed, because CloudFront is likely not causing the issue that we're trying to debug.

PR #58, opened 2024-02-20
Set runAsGroup and fsGroup on the deployment.

EarthlingDavey:
Potentially, www-user (82) cannot write to the mounted volume /var/www/html/public/app/uploads . @see https://stackoverflow.com/a/57917406/6671505


EarthlingDavey:
On hold. I'm reviewing the changes locally with kind.

wilson1000-MoJ:
🤔 hmm, I'm thinking this can't be the case. I haven't checked but I can't remember defining `fsGroup`, yet when we deployed the application [we could upload and edit images](https://mojdt.slack.com/archives/D06DVH7RVBQ/p1708354875576189).

EarthlingDavey:
I see, yeah, let's close this PR without merging.

PR #62, opened 2024-02-26
Add sentry

EarthlingDavey:


config/environments/development.php
```diff
@@ -19,3 +19,6 @@
 if (env('WP_OFFLOAD_MEDIA_PRESET')) {
     Config::define('WP_OFFLOAD_MEDIA_PRESET', env('WP_OFFLOAD_MEDIA_PRESET'));
 }
+
+const SENTRY_TRACES_SAMPLE_RATE = 1.0;
+const SENTRY_PROFILE_SAMPLE_RATE = 1.0;
```
wilson1000-MoJ: Should be 
```
Config::define('SENTRY_TRACES_SAMPLE_RATE', 1.0);
```
To prevent constant clashes in other files. Unless you'd like to change the values on different platforms I would put these in application.php - the catch all configuration (prod) file

deploy/demo/deployment.tpl
```diff
@@ -70,6 +70,11 @@ spec:
                 secretKeyRef:
                   name: rds-output
                   key: database_password
+            - name: SENTRY_DSN
+              valueFrom:
+                secretKeyRef:
+                  name: sentry-output
+                  key: sentry-dsn
```
wilson1000-MoJ: This code assumes Cloud Platform created the secret and Kubernetes will pass it through. 
We can add this to `secret.tpl` so that it can pick up the GitHub variable, like this:

```
apiVersion: v1
kind: Secret
metadata:
  name: justice-gov-uk-dev-secrets
type: Opaque
stringData:
  SENTRY_DSN: "${SENTRY_DSN}"
  GOV_NOTIFY_API_KEY: "${GOV_NOTIFY_API_KEY}"
  AUTH_KEY: "${AUTH_KEY}"
  ......
```

wilson1000-MoJ (approved):

PR #65, opened 2024-02-27
Add tunnelling to php-fpm container.

EarthlingDavey:
This PR should allow us to debug and explore our issue with wp-offload-media. 

Debugging should be easier with a local codebase. Connecting to a pod and using live server-role credentials.

Please see the code comments for the methodology.


EarthlingDavey:
This branch was was a potential security vuln, and the underlying issue with wp-offload-media was resolved via another approach.

PR #74, opened 2024-03-01
Post meta for sidebar fields

EarthlingDavey:
`justice/inc/simple-guten-fields` and `justice/src/components/simple-guten-fields` are an extensible plugin used for showing custom fields on posts/pages.

It is based upon this post, https://bebroide.medium.com/how-to-easily-develop-with-react-your-own-custom-fields-within-gutenberg-wordpress-editor-b868c1e193a9 with customisation for our use case.

Our fields for the above plugin are defined in `justice/inc/post-meta/constants.php`.

Current fields include toggles for sidebars, and entries for some sidebars e.g. related content.


docker-compose.yml
```diff
@@ -13,6 +13,7 @@ services:
             context: .
             target: dev
         volumes:
+            - dist:/var/www/html/public/app/themes/justice/dist
```
wilson1000, line 16: Why does this need access to this directory?

wilson1000 (approved):

docker-compose.yml
EarthlingDavey: As part of the js build: laravel mix + @tinypixelco/laravel-mix-wp-blocks will create n asset file e.g. `dist/block-editor.asset.php` it contains dependencies and a version number. These are then used is then used in `wp_register_script`.

In docker-compose, this is above the `.:/var/www/html` because the order of them makes a difference.

PR #76, opened 2024-03-01
fix for missing asset.php file

EarthlingDavey:
- webpack to copy asset.php files to `dist/php`
- update the path in php application
- Dockerfile
    - Move `assets-build` above `build-fpm`
    - minimise files copied over to `assets-build` from the codebase
    - build-fpm to get `dist/php` from `assets-build`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,6 +72,18 @@ RUN mkdir -p ./vendor-assets && \
     find public/ -regex "public\/${regex_path}.*\.${regex_files}" -exec cp --parent "{}" vendor-assets/  \;
 
 
+###
+
+
+FROM node:20 AS assets-build
+
+WORKDIR /code/public/app/themes/justice
+COPY ./public/app/themes/justice ./
+
+RUN npm ci
+RUN npm run production
+RUN rm -rf node_modules
+
 
 ###
 
@@ -79,7 +91,8 @@ RUN mkdir -p ./vendor-assets && \
 FROM base-fpm AS build-fpm
 
 WORKDIR /var/www/html
-COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html /var/www/html
+COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html .
+COPY --from=assets-build       --chown=www-data:www-data /code/public/app/themes/justice/dist/php ./public/app/themes/justice/dist/php
 
 # non-root
 USER 82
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -72,14 +72,27 @@ RUN mkdir -p ./vendor-assets && \
     find public/ -regex "public\/${regex_path}.*\.${regex_files}" -exec cp --parent "{}" vendor-assets/  \;
 
 
+###
+
+
+FROM node:20 AS assets-build
+
+WORKDIR /code/public/app/themes/justice
```
wilson1000, line 80: `code/` is a directory we create, using WORKDIR, to isolate the compilation step.

If you take a look in the `docker-compose` node service, you'll see we only take what we need to compile assets - instead of `code`, we call the directory `node`. We could do this here too, maybe?

wilson1000: What I mean is, there is no need, other than getting needed files, to recreate the directory structure.

Dockerfile
```diff
@@ -72,14 +72,27 @@ RUN mkdir -p ./vendor-assets && \
...
+COPY ./public/app/themes/justice ./
+
```
wilson1000, line 82: You could copy all of this into `code/`

Dockerfile
```diff
@@ -72,14 +72,27 @@ RUN mkdir -p ./vendor-assets && \
...
 
 ###
 
 
 FROM base-fpm AS build-fpm
 
 WORKDIR /var/www/html
-COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html /var/www/html
+COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html .
+COPY --from=assets-build       --chown=www-data:www-data /code/public/app/themes/justice/dist/php ./public/app/themes/justice/dist/php
```
wilson1000, line 95: Then this would be:

```
COPY --from=assets-build --chown=www-data:www-data /code/dist/php ./public/app/themes/justice/dist/php
```

Dockerfile
EarthlingDavey: Good spot!

Dockerfile
EarthlingDavey: Thanks, I've done this in the new commit. Except the folder name is node, to match the folder name in docker-compose.yml

wilson1000 (approved):
Looks good 👍

PR #86, opened 2024-03-08
Tests unit

EarthlingDavey:
- Add unit tests for some classes.
- Add file watcher script, to run tests on file changes.
- Document in readme file.
- Unit tests run in the php-test.yml workflow.

---

There is a different test branch that has integration and e2e tests. They're not fully working.


EarthlingDavey:
When [this issue](https://github.com/10up/wp_mock/issues/243) is resolved. Set the version of 10up/wp-mock to ^1.0.1

config/application.php
```diff
@@ -81,16 +81,15 @@
...
+|| getenv('WPBROWSER_HOST_REQUEST');
+
+// Set the table prefix based on the request origin.
+$table_prefix =  $is_test_request ? 'test_' : (env('DB_PREFIX') ?: 'wp_');
 
-    Config::define('DB_NAME', substr($dsn->path, 1));
-    Config::define('DB_USER', $dsn->user);
-    Config::define('DB_PASSWORD', $dsn->pass ?? null);
-    Config::define('DB_HOST', isset($dsn->port) ? "{$dsn->host}:{$dsn->port}" : $dsn->host);
-}
```
wilson1000, line 93: I can't see a replacement here yet, a URL has been defined in the `.env.example` file. 
Is this important?

spec/Support/Data/.gitkeep
wilson1000, line 1: Are other files like this called `.keep`? 🤔

spec/Support/_generated/.gitignore
wilson1000, line 1: We would usually maintain one `.gitignore` - do tests require this file or can we move the path to the root `.gitignore` file?

```
spec/Support/_generated/*
```

spec/_output/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
wilson1000, line 1: Same comment on this one; can we move the rule to the root `.gitignore` file? 
This will help us reduce maintenance complexity.

spec/Unit/mock-template-part.php
```diff
@@ -0,0 +1 @@
+<h1><?php echo $args['title']; ?></h1>
```
wilson1000, line 1: Quick discussion question, if I may 😄  ...

```
<?php echo $args['title']; ?>
```

... over ...

```
<?= $args['title'] ?>
```

We have a more verbose usage or minimal PHP code usage. 
Should we use minimal code as we are injecting a variable in HTML? Or does verbose output make more sense?

wilson1000 (dismissed):
Some great work here @EarthlingDavey well done on the Unit Tests!!

I'll approve. Do take a look at the comments and let's have a discussion 👍

config/application.php
EarthlingDavey: I've checked it looks like we're not using `DATABASE_URL` for local or deployments.

I think it would be for if we use a connection string like `mysql://<dbhost>:<dbport>/<dbname>`.

My thoughts are we probably won't start using `DATABASE_URL` so let's remove it from the codebase.

spec/Support/Data/.gitkeep
EarthlingDavey: Cheers, will `.keep` in line with the other files 😉

spec/Support/_generated/.gitignore
EarthlingDavey: Good spot, they were there from installing `lucatume/wp-browser`. I'll move to the root `.gitignore` :)

EarthlingDavey: Good spot, they were there from installing `lucatume/wp-browser`. I'll move to the root `.gitignore` :)

spec/Unit/mock-template-part.php
EarthlingDavey: I prefer `<?= $args['title'] ?>` in templates. `echo` is just a habit.

Are you good for me to change this one? And any others in a different PR?

wilson1000 (approved):
Nice 💯

PR #88, opened 2024-03-13
Add nginx proxy-body-size annotation to allow for large file uploads.

EarthlingDavey:
Updates the deployments to allow file uploads < 200MB

- Docs: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#proxy-body-size
- Chat: https://mojdt.slack.com/archives/C57UPMZLY/p1647342791594759?thread_ts=1647342045.542929&cid=C57UPMZLY
- e.g. https://github.com/ministryofjustice/evidence-library/blob/main/src/kubectl_deploy/test-env/ingress.yaml

### DIFF
--- a/deploy/staging/ingress.yml
+++ b/deploy/staging/ingress.yml
@@ -9,6 +9,7 @@ metadata:
     nginx.ingress.kubernetes.io/auth-type: basic
     nginx.ingress.kubernetes.io/auth-secret: basic-auth-secret
     nginx.ingress.kubernetes.io/auth-realm: 'Staging User | Authentication Required'
+    nginx.ingress.kubernetes.io/proxy-body-size: "200m"
     nginx.ingress.kubernetes.io/server-snippet: |
       location = /health {
         auth_basic off;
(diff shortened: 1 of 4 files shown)


deploy/staging/ingress.yml
```diff
@@ -9,6 +9,7 @@ metadata:
     nginx.ingress.kubernetes.io/auth-type: basic
     nginx.ingress.kubernetes.io/auth-secret: basic-auth-secret
     nginx.ingress.kubernetes.io/auth-realm: 'Staging User | Authentication Required'
+    nginx.ingress.kubernetes.io/proxy-body-size: "200m"
```
wilson1000, line 12: Just thinking. Will our PHP-FPM allow this size of upload?

wilson1000 (approved):

deploy/staging/ingress.yml
EarthlingDavey: It works locally. 
Famous last words!

PR #91, opened 2024-03-14
Multiple changes around document post type and importing.

EarthlingDavey:


EarthlingDavey:
The build was failing at composer install.

It was caused by composer running without a lock file, then running with a lock file, and causing a downgrade. And not being able to delete the files of the newest version. We need to review if we should be doing it like that.

For now, it's fixed by updating `wpackagist-plugin/debug-bar` to the newest version: `1.1.5`

public/app/themes/justice/inc/documents.php
```diff
@@ -296,4 +360,89 @@ public function addRewriteRules($rules)
...
+     * - setUploadSizeLimit
+     */
+
+
+    /**
+     * isValidSlug
+     * Prevent all single posts, of any post type, from using the document(s) slug.
+     * This prevents a conflict, where nested pages may make a url like /documents/my-document
+     */
+
```
EmilyHazlehurst, line 378: @EarthlingDavey this is more of an observation than an issue, but is it worth using PHPDoc comments to describe functions? Probably more for a team discussion around code standards generally.

```
/**
* Prevent all single posts, of any post type, from using the document(s) slug.
*
* @param bool $bad_slug Whether the slug is bad or not
* @param string $slug The document slug
* @return bool The sum of the two numbers.
*/
```

https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/

EmilyHazlehurst (approved):

Issue #78, opened 2024-03-05
Format code with PSR


EarthlingDavey:
As @wilson1000-MoJ suggested in this review. https://github.com/ministryofjustice/justice-gov-uk/pull/77#discussion_r1512622834 

It's important to use a standard for code readability, and minimise that chance of merge conflicts.

Let's revisit.

EarthlingDavey:
@EmilyHazlehurst has raised an important point about using PHPDoc comments to describe functions. https://github.com/ministryofjustice/justice-gov-uk/pull/91#discussion_r1524881714

e.g. 

```php
/**
* Prevent all single posts, of any post type, from using the document(s) slug.
*
* @param bool $bad_slug Whether the slug is bad or not
* @param string $slug The document slug
* @return bool The sum of the two numbers.
*/
```

https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/

+1 from me. Let's have a team discussion.

EarthlingDavey:
Adding to this, we could potentially:

- phpcs format on save.
- pre commit hook to run phpcs and/or unit tests etc.
- CI build for generating browsable docs. [WP use](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/) [phpDocumentor](https://phpdoc.org/3.0/), I've had a trial and it looks really nice!

EarthlingDavey:
Further research PSR-4.

Implementation would mean adding `autoloader` section of `composer.json`. 
This will require `build-fpm-composer` has access to php source code in the `Dockerfile`.

Resources:
- https://gitlab.com/rawkode/php-examples/-/blob/main/docker-and-kubernetes/pingcrm/composer.json?ref_type=heads#L42
- https://www.youtube.com/live/z_LWJPLhD6g?t=4475
- https://getcomposer.org/doc/04-schema.md#autoload
- https://www.php-fig.org/psr/psr-4/
- https://dlxplugins.com/tutorials/creating-a-psr-4-autoloading-wordpress-plugin/
- https://medium.com/write-better-wordpress-code/psr-4-autoloading-in-plugins-and-themes-720f459f4df4
- https://www.mrkwp.com/guide/convert-shortcode-to-block-wordpress/using-composer-and-psr-4/

PR #93, opened 2024-03-18
Optimise - remove unused core functions. Document filtering & columns.

EarthlingDavey:
- Remove unused WordPress core functions, using filters and actions.
- Filtering and columns on document post type. Useful for identifying documents without attachments or revisions.
- Trial phpdocs in `/inc/documents` folder.

### DIFF
--- a/public/app/themes/justice/functions.php
+++ b/public/app/themes/justice/functions.php
@@ -18,6 +18,7 @@ if (Config::get('WP_OFFLOAD_MEDIA_PRESET') === 'minio') {
 require_once 'inc/admin.php';
 require_once 'inc/block-editor.php';
 require_once 'inc/breadcrumbs.php';
+require_once 'inc/core.php';
 require_once 'inc/debug.php';
 require_once 'inc/disable-comments.php';
 require_once 'inc/documents/documents.php';
@@ -25,7 +26,6 @@ require_once 'inc/dynamic-menu.php';
 require_once 'inc/errors.php';
 require_once 'inc/layout.php';
 require_once 'inc/mail.php';
-require_once 'inc/optimise.php';
 require_once 'inc/post-meta/post-meta.php';
 require_once 'inc/simple-guten-fields/simple-guten-fields.php';
 require_once 'inc/taxonomies.php';
@@ -38,9 +38,9 @@ if (getenv('WP_ENV') === 'development') {
 
 new Justice\Admin();
 new Justice\Comments();
+new Justice\Core();
 new Justice\Documents();
 new Justice\Layout();
-new Justice\Optimise();
 new Justice\SimpleGutenFields();
 
 $block_editor = new Justice\BlockEditor();
--- a/public/app/themes/justice/functions.php
+++ b/public/app/themes/justice/functions.php
@@ -20,7 +20,7 @@ require_once 'inc/block-editor.php';
 require_once 'inc/breadcrumbs.php';
 require_once 'inc/debug.php';
 require_once 'inc/disable-comments.php';
-require_once 'inc/documents.php';
+require_once 'inc/documents/documents.php';
 require_once 'inc/dynamic-menu.php';
 require_once 'inc/errors.php';
 require_once 'inc/layout.php';
--- a/public/app/themes/justice/functions.php
+++ b/public/app/themes/justice/functions.php
@@ -29,6 +29,7 @@ require_once 'inc/optimise.php';
 require_once 'inc/post-meta/post-meta.php';
 require_once 'inc/simple-guten-fields/simple-guten-fields.php';
 require_once 'inc/taxonomies.php';
+require_once 'inc/utils.php';
 
 if (getenv('WP_ENV') === 'development') {
     $debug = new Justice\Debug();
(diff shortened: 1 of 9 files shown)


EarthlingDavey:
Composer install is breaking the build so this commit also includes a fix.

The current dockerfile:

```
COPY ./composer.json /var/www/html/composer.json
RUN composer install --no-dev --no-scripts --no-autoloader

COPY . .
RUN composer install --no-dev
```

- Installs the newest version for any package when the version is set to '*', e.g. `debug-bar`.
- Then COPY . . will copy the composer.lock file.
- The 2nd composer install will install the locked versions. Which may be behind latest
- Sometimes the 2nd install will need to remove a file, and it fails at this step.

The issue... can be seen on this workflow https://github.com/ministryofjustice/justice-gov-uk/actions/runs/8325274173/job/22778565139 or locally with the command `docker build . --target=build-nginx` 

The change.

- Copy composer.json and composer.lock and run composer install once.
- Don't run `COPY . .` in `build-fpm-composer` and adjust `build-fpm` & `build-nginx` accordingly.

Dockerfile
```diff
@@ -59,10 +59,7 @@ COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
 # non-root
 USER 82
 
-COPY ./composer.json /var/www/html/composer.json
-RUN composer install --no-dev --no-scripts --no-autoloader
-
-COPY . .
+COPY ./composer.json ./composer.lock /var/www/html/
```
wilson1000, line 62: This is a production build, we will need `--no-dev` - unless tested

Dockerfile
```diff
@@ -125,5 +125,5 @@ COPY --from=assets-build /node/dist      /var/www/html/public/app/themes/justice
 
 # Only take what Nginx needs (current configuration)
 COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html/vendor-assets /var/www/html/
-COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html/public/index.php /var/www/html/public/index.php
+COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html/public/wp/index.php /var/www/html/public/index.php
```
wilson1000, line 128: This looks like a probable typo. 
If we don't have an index.php file in the root of the service Nginx will fail.

wilson1000: Ok, just saw that we're picking the file from the WP core. We have a modified file in public here, do we not need this anymore?

EarthlingDavey: Sorry, my bad, I was too hasty changing this. I thought it was the same as `public/index.php`. I've reverted this.

Dockerfile
```diff
@@ -59,11 +59,9 @@ COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
 # non-root
 USER 82
 
-COPY ./composer.json /var/www/html/composer.json
-RUN composer install --no-dev --no-scripts --no-autoloader
+COPY ./composer.json ./composer.lock /var/www/html/
```
EarthlingDavey, line 62: Methodology: 1. Add lock here because I don't think our codebase is in the best position for uniing w/o lock at the moment.

Dockerfile
```diff
@@ -59,11 +59,9 @@ COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
...
+RUN composer install --no-dev --no-autoloader
 
-COPY . .
```
EarthlingDavey, line 65: Methodology: 2. Remove `COPY . .` because having access to the php source is only beneficial if the files are defined in autoloaded in composer.json. I've created this issue to re-visit. https://github.com/ministryofjustice/justice-gov-uk/issues/78#issuecomment-2003998333

Dockerfile
```diff
@@ -96,7 +94,10 @@ RUN rm -rf node_modules
 FROM base-fpm AS build-fpm
 
 WORKDIR /var/www/html
-COPY --from=build-fpm-composer --chown=www-data:www-data /var/www/html  .
+COPY --chown=www-data:www-data ./config ./config
```
EarthlingDavey, line 97: Methodology: 4. Since we're not doing `COPY . .` , selectively copy folders here.

Dockerfile
```diff
@@ -119,11 +120,13 @@ FROM base-nginx AS build-nginx
 COPY deploy/config/php-fpm.conf /etc/nginx/php-fpm.conf
 COPY deploy/config/server.conf /etc/nginx/conf.d/default.conf
 
+# WordPress view bootstrapper
+COPY public/index.php /var/www/html/public/index.php
```
EarthlingDavey, line 124: Methodology: 5. Since we're not doing COPY . . , copy index.php from host instead.

EarthlingDavey (commented):
Added annotations to Dockerfile changes.

wilson1000 (approved):

Dockerfile
wilson1000: This is quite a confusing subject. Thanks for creating the issue 💯

wilson1000 (approved):

Dockerfile
```diff
@@ -59,11 +59,9 @@ COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
 # non-root
 USER 82
 
-COPY ./composer.json /var/www/html/composer.json
-RUN composer install --no-dev --no-scripts --no-autoloader
+COPY ./composer.json ./composer.lock /var/www/html/
+RUN composer install --no-dev --no-autoloader
```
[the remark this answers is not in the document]
EarthlingDavey: Sorted, cheers :)

wilson1000 (approved):

PR #98, opened 2024-04-22
Templates for search results and relevanssi premium.

EarthlingDavey:


public/app/themes/justice/inc/search.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        if (!is_admin() && empty($search) && $q->is_search() && $q->is_main_query()) {
+            // Return an empty search query to prevent any results from being returned.
+            $search .= " AND 0=1 ";
+        }
+
+        return $search;
+    }
+
+    // TODO: Redirect old query string search URLs to the new search page.
```
wilson1000, line 141: Is this a work in progress?
Are you happy for this to remain for now?

wilson1000 (approved):
Great work @EarthlingDavey 👍

public/app/themes/justice/inc/search.php
EarthlingDavey: Good spot, yes, it's one of the next things on my list.

PR #115, opened 2024-05-01
CDPT-1544 Schedule page changes.

EarthlingDavey:
@EmilyHazlehurst this is the mostly complete solution for having scheduled page revisions.

Todo:

- Use css or hooks to remove the branding from the plugin.
- Remove unnecessary items from the sidebar like Upgrade To Pro.
- Move the Dockerfile edits into the wordpress base image
- Have a thorough test of the plugin functionality. I managed to hit a 'headers already sent error' but didn't get as far as debugging it.
- The end users will have an 'editor' role so it must work for them.


public/app/themes/justice/inc/admin.php
```diff
@@ -36,4 +37,41 @@ public static function removeCustomizer()
...
+                 footer {
+                   display: none;
+                 }  
+               }
+               .ppseries-settings-body-content {
+                 width: 100% !important;
+               }
+               .rvy-creation-ui {
+                 display: none;
+               }
```
EmilyHazlehurst, line 74: ^ I'm hiding the UI elements in the visual editor because the preview doesn't work (it seems to be a bug in the plugin).

There are buttons in the admin interface that do work so this shouldn't be an issue.

public/app/themes/justice/inc/admin.php
```diff
@@ -36,4 +37,41 @@ public static function removeCustomizer()
         $customizer_url = add_query_arg('return', urlencode(remove_query_arg(wp_removable_query_args(), wp_unslash($_SERVER['REQUEST_URI']))), 'customize.php');
         remove_submenu_page('themes.php', $customizer_url);
     }
+
+    public static function removePublishPressUpsellBranding()
+    {
+        echo '<style>
+               .toplevel_page_revisionary-q {
+                 li:last-of-type {
+                   display: none;
```
EarthlingDavey, line 46: This is too broad sweeping - it targets every list on the Revision Queue page. e.g. There's a 'Collapse menu' button at the bottom of the sidebar, this css rule set's that to display none.

This rule doesn't target other pages, i.e. Revisions Archive or Settings page of this plugin.

Perhaps this link could be targeted by an attribute?

public/app/themes/justice/inc/admin.php
```diff
@@ -36,4 +37,41 @@ public static function removeCustomizer()
...
+                 }
+                 footer {
+                   display: none;
+                 }  
+               }
+               .ppseries-settings-body-content {
+                 width: 100% !important;
+               }
+               .rvy-creation-ui {
```
EarthlingDavey, line 72: I think removing the UI for Submit or Approve has the unintended consequence of the revisions sitting in the Revision Queue and being in the Not Submitted state.

EarthlingDavey (commented):
Hey @EmilyHazlehurst , this looks good mostly good :)

Its a nice touch making the settings 100% width to account for the removed element.

I've added a few comments for you to check before we merge this in.

If they need clarifying or if you hit a wall with any of them give me a shout :)

public/app/themes/justice/inc/admin.php
EarthlingDavey: I see, do you mind sharing steps to reproduce? It might be worth fixing the bug.

See [my comment](https://github.com/ministryofjustice/justice-gov-uk/pull/115#discussion_r1588955479) about the unintended consequence of this rule.

public/app/themes/justice/inc/admin.php
```diff
@@ -36,4 +37,41 @@ public static function removeCustomizer()
...
+                 li:last-of-type {
+                   display: none;
+                 }
+               }
+               .pp-version-notice-bold-purple {
+                 display: none;
+               }
+               .pressshack-admin-wrapper {
+                 a {
+                  color: #2271b1;
```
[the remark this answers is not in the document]
EmilyHazlehurst: That was intentional - I was making the branded publishpress purple links match the default wordpress blue as the purple they use fails WCAG contrast guidelines.

public/app/themes/justice/inc/admin.php
EmilyHazlehurst: I'm not sure it makes a difference to the flow? I'm not removing submit/approve here, only Preview/Edit.

The flow in both cases is:
1. Make a new revision
2. Preview revision
3. Submit and approve from the preview/edit page

public/app/themes/justice/inc/admin.php
EmilyHazlehurst: Sure, steps to reproduce are: 

1. Edit a page and click New Revision.
Screenshot 2024-05-03 at 14 03 05
2. Click the Preview or Edit button that appears.
Screenshot 2024-05-03 at 14 04 33
3. You will be redirected to the wrong page and see either a white screen or the 'Headers already sent' error message.

The preview and edit buttons in the revision queue work correctly.

public/app/themes/justice/inc/admin.php
```diff
@@ -36,4 +37,41 @@ public static function removeCustomizer()
         $customizer_url = add_query_arg('return', urlencode(remove_query_arg(wp_removable_query_args(), wp_unslash($_SERVER['REQUEST_URI']))), 'customize.php');
         remove_submenu_page('themes.php', $customizer_url);
     }
+
+    public static function removePublishPressUpsellBranding()
```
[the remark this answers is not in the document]
EmilyHazlehurst: Agree! I wasn't sure what the standard was with Wordpress and all of the examples used inline, I've moved it into admin.scss :)

public/app/themes/justice/inc/admin.php
EmilyHazlehurst: Thanks, I've changed it to target the link with the matching href.

public/app/themes/justice/inc/admin.php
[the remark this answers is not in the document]
EarthlingDavey: Nice! That's one thing with WP, sometimes the examples (especially some answers on StackOverflow etc.) won't be best practice.

public/app/themes/justice/src/sass/admin.scss
```diff
@@ -0,0 +1,33 @@
+.toplevel_page_revisionary-q {
+    a[href="admin.php?page=revisionary-pro"] {
```
EarthlingDavey, line 2: I think targeting the href is the best way to do this.

Can you check it works locally?

public/app/themes/justice/inc/admin.php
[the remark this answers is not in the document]
EarthlingDavey: oh wow, nice catch. that's pretty cheeky of the plugin author 🐵

Do you mind commenting the reasoning, and, checking the line indents.

public/app/themes/justice/inc/admin.php
EarthlingDavey: When I check locally, it's removing the Submit / Approve UI.

image

To reproduce, browse the frontend and click 'New Revision' in the top bar. Or in admin, All Pages -> hover a page -> New Revision.

I wonder if we could fix the bug, maybe there's a conflict with the theme or another plugin.

public/app/themes/justice/inc/admin.php
EarthlingDavey: Hey @EmilyHazlehurst , I had a poke around and found the reason for the bug.

The plugin was sending redirect headers after the html had started rendering.

I've changed the hooks to ones that fire before the html starts to render, would you mind copy pasting the bit of code in this PR to your local just to verify that it works. Thanks :)

https://github.com/publishpress/PublishPress-Revisions/pull/1238

EmilyHazlehurst: Thanks @EarthlingDavey, that's now in the plugin and works perfectly when updated :)

public/app/themes/justice/inc/admin.php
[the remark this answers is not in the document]
EmilyHazlehurst: I've added comments and fixed the indents

public/app/themes/justice/inc/admin.php
EmilyHazlehurst: Fixed with the latest version of the plugin

public/app/themes/justice/src/sass/admin.scss
EmilyHazlehurst: Yes, this works locally.

wilson1000:
Hey @EmilyHazlehurst 
Some CSS suggestions from our conversation:

Front end to being the colour of the background in line with design:
```scss
.rvy_preview_msgspan {
  background-color: #373737 !important;
}
```

In the admin to remove the flash from the CSS hiding the PRO link
```scss
ul.wp-submenu-wrap {
  li:last-child {
    display: none !important;
  }
}
```

EarthlingDavey:
@wilson1000 I think the last-child will be too wide sweeping. If theres a flash because the plugin is using js to cheekily change the url, it might be better to target both urls, something like:

```
a[href="admin.php?page=revisionary-pro"],
a[href="https://publishpress.com/links/revisions-menu"] {
  display: none !important;
}
```

public/app/themes/justice/src/sass/admin.scss
```diff
@@ -45,3 +47,8 @@
 div.cms_tpv_promo_above_wrapper {
     display: none !important;
 }
+
+// Recolour edit banner to better fit the theme
+.rvy_preview_msgspan {
+  background-color: #373737 !important;
+}
```
wilson1000, line 49: Love this @EmilyHazlehurst - to be certain; will this file display on the front end if it's an admin style?

wilson1000:
> @wilson1000 I think the last-child will be too wide sweeping. If theres a flash because the plugin is using js to cheekily change the url, it might be better to target both urls, something like:
> 
> ```
> a[href="admin.php?page=revisionary-pro"],
> a[href="https://publishpress.com/links/revisions-menu"] {
>   display: none !important;
> }
> ```

I've tested locally; the selectors are too direct and cumbersome. For example; the DOM is under too much pressure using that selector. In addition, the CSS rule still overrides and JS additions.

EarthlingDavey:
It is always a bit risky using wide spanning selectors like that. For example you can see it does target some other menu items.

image

I wouldn't be too worried about dom pressure from 2 anchor selectors, I've seen worse.

Anyways, I'll hand the review over to you @wilson1000 because otherwise we might be going round in circles.

Give me a shout if you want me to dip back in.

public/app/themes/justice/src/sass/admin.scss
EmilyHazlehurst: @wilson1000 this was moved to the app.scss file, you must have commented as it was being pushed. That should be working now :)

wilson1000:
Hey @EmilyHazlehurst & @EarthlingDavey - will this be ok?

This will isolate the rule, also on the plugin page and remove the dependency on URLs

```sass
> ul.wp-submenu-wrap {
  li:last-child {
    display: none !important;
  }
}
```

EarthlingDavey:
@wilson1000 that looks good for admins. But the nuance of editors not having 'Settings' or 'Upgrade to Pro', means that for them 'Revision Archive' will be hidden.

public/app/themes/justice/src/sass/admin.scss
```diff
@@ -1,3 +1,49 @@
+// Styling for the PublishPress Revisions plugin
+
+// Hide the upsell menu link
+.toplevel_page_revisionary-q {
+  ul.wp-submenu-wrap {
```
wilson1000, line 5: Hey @EmilyHazlehurst 😄 
Davey has put a brilliant case forward for using URL selection instead. I'm sorry for muddying the waters here with my suggestion.

I was wrong about the DOM load problem I thought we had. JS was changing the URL, causing the page's jump effect. I've enjoyed the conversation around this stuff.

Thanks to both you and @EarthlingDavey for the opportunity 👍

EmilyHazlehurst: Thanks @wilson1000 @EarthlingDavey, I think that should be everything but let me know if there's anything else :)

wilson1000 (approved):
Hey Emily, I've noticed the linting applied removed standards spacing. 
Let's not get into it here. Let's chat about standards and agree on what we want to follow once the migrations are done.

I've approved this and I'm looking forward to the release 😄

PR #130, opened 2024-05-08
CDPT-1562: Issue with dropdown

EmilyHazlehurst:
- Increase the page limit on the dropdown editor


EarthlingDavey (commented):
Hey @EmilyHazlehurst this looks nice!

Just a couple of minors

Since you suggested writing phpdoc style comments I've been trying to do that. I think it's just unfortunate that where you've been writing your functions is where I hadn't used phpdoc style comments yet 🙃

So... adding arg type and return type of array would be a nice touch.

And, is there a css property that we could set in editor.scss to make the dropdown select list higher? I think this would vastly improve the editor experience.

You see where the composer test failed, you can run this locally with `make bash` then `composer test:full`, and fix some errors automagically with `composer test:fix`.

BTW thanks for picking up these tickets 🙌

public/app/themes/justice/inc/admin.php
```diff
@@ -36,4 +37,14 @@ public static function removeCustomizer()
         $customizer_url = add_query_arg('return', urlencode(remove_query_arg(wp_removable_query_args(), wp_unslash($_SERVER['REQUEST_URI']))), 'customize.php');
         remove_submenu_page('themes.php', $customizer_url);
     }
+    
+    /**
+     * Increase the dropdown limit on the block editor parent page field
+     */
+    function increaseDropdownLimit( $args ) {
+        if ( is_user_logged_in() ) {
```
EarthlingDavey, line 45: Would it be possible to add another check here to make the page increase more targeted?

For example you could access the $request parameter, and use `$request->get_query_params()['context'], it will equal `edit` in this case.

This is a good resource for getting info out of the request.

https://developer.wordpress.org/reference/classes/wp_rest_request/

I think it could help prevent a confusing bug further down the line where we're wondering why our other ajax requests are maxing out at 2000 😆

EmilyHazlehurst:
@EarthlingDavey thanks for the feedback :) 

Please could you clarify: 

> And, is there a css property that we could set in editor.scss to make the dropdown select list higher? I think this would vastly improve the editor experience.

Do you mean move it up the page, or something else?

EarthlingDavey:
> Do you mean move it up the page, or something else?

Oops, sorry... I mean increase the hight so editors can scroll through and see maybe a half a dozen to a dozen at a time. 

It might be easy enough to target the element with css and over-ride the hight.

EmilyHazlehurst:
> > Do you mean move it up the page, or something else?
> 
> Oops, sorry... I mean increase the hight so editors can scroll through and see maybe a half a dozen to a dozen at a time.
> 
> It might be easy enough to target the element with css and over-ride the hight.

Thanks @EarthlingDavey, I've updated that. I've tripled the box height; any more and it flickers quite badly when the list updates.

public/app/themes/justice/inc/admin.php
```diff
@@ -36,4 +39,23 @@ public static function removeCustomizer()
...
+     *
+     * @param array $args An array of arguments for WP_Query
+     * @param WP_REST_Request $request The REST API request
+     *
+     * @return array $args The modified WP_Query array
+     */
+    public static function increaseDropdownLimit($args, $request)
+    {
+        if (is_user_logged_in()) {
+            if ($request->get_query_params()['context'] === 'edit') {
```
EarthlingDavey, line 54: Super minor one these could be on the same line.

public/app/themes/justice/src/sass/editor.scss
```diff
@@ -1,2 +1,7 @@
 @import "../components/search/editor";
 @import "../components/simple-definition-list-blocks/editor";
+
+// Triple the height of the suggestion list for the combobox component
+.components-form-token-field__suggestions-list {
+  max-height: 384px !important;
+}
```
EarthlingDavey, line 7: With an aim for consistency, could you add a line break at the end of the file.

EarthlingDavey:
> I've tripled the box height

Looks good to me. I think the editors will appreciate it! I've added a final couple of minor points.

public/app/themes/justice/inc/admin.php
wilson1000: Standards would suggest different lines @EarthlingDavey, for readability purposes.
We follow PSR-12 to coding standards, sorry if I'd not ever mentioned that 😵‍💫 
https://www.php-fig.org/psr/psr-12/

public/app/themes/justice/src/sass/editor.scss
EmilyHazlehurst: Thanks @EarthlingDavey that should be fixed :)

public/app/themes/justice/inc/admin.php
EarthlingDavey: @wilson1000 I was thinking change it to one if statement with `&&` all on one line. 

Admittedly, I'm 100% not following psr-12 but I think that's allowed.

> Expressions in parentheses MAY be split across multiple lines

wilson1000 (approved):
Looks good @EmilyHazlehurst - well done 🎉

PR #132, opened 2024-05-08
CDPT-1076 Add theme support for title-tag.

EarthlingDavey:
The current theme had issues setting the `<title>` element on the homepage.

Adding theme support for `title-tag` is the recommended approach for fixing the issue. 

Reference:

- https://developer.wordpress.org/reference/functions/add_theme_support/
- https://stackoverflow.com/a/37223205/6671505
- https://wordpress.stackexchange.com/questions/305353/cant-change-the-title-tag-with-wp-title-filter


wilson1000 (approved):
Looks good 😄

PR #137, opened 2024-05-13
Handle requests to the application host, by sending them to a loopback url.

EarthlingDavey:
This PR sets a config of `WP_LOOPBACK` to one that is defined in the environment variables.

If there's a WP request to `home_url` then this is replaced with `WP_LOOPBACK`. 

As `WP_LOOPBACK` can be the nginx container, we don't need ssl, or basic auth.

Some un-needed local config has also been removed/reverted.

### DIFF
--- a/deploy/config/local/server.conf
+++ b/deploy/config/local/server.conf
@@ -15,7 +15,6 @@ upstream php {
 
 server {
     listen  8080 default_server; # For default requests.
-    listen  80;                  # For localhost loopback only.
     server_name localhost;
 
     root /var/www/html/public;
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,8 +18,6 @@ services:
         depends_on:
             - mariadb
             - minio
-        links:
-          - "nginx:${SERVER_NAME}"
 
     nginx:
       build:
@@ -39,6 +37,7 @@ services:
         - "8080:8080"
       depends_on:
         - node
+        - php-fpm
 
     node:
         image: node:20
--- a/public/app/themes/justice/inc/core.php
+++ b/public/app/themes/justice/inc/core.php
@@ -2,12 +2,13 @@
 
 namespace MOJ\Justice;
 
-if (!defined('ABSPATH')) {
-    exit;
-}
+use WP_Error;
+use Roots\WPConfig\Config;
+
+defined('ABSPATH') || exit;
 
 /**
- * Actions and filters related to removing unused features from WordPress core.
+ * Actions and filters related to WordPress core.
  */
 
 class Core
@@ -28,6 +29,8 @@ class Core
         add_filter('should_load_remote_block_patterns', '__return_false');
         // Avoids unnecessary transient entry in the database, by returning an empty array.
         add_filter('translations_api', fn () => []);
+        // Handle loopback requests.
+        add_filter('pre_http_request', [$this, 'handleLoopbackRequests'], 10, 3);
     }
 
     /**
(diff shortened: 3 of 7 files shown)


docker-compose.yml
```diff
@@ -39,6 +37,7 @@ services:
         - "8080:8080"
       depends_on:
         - node
+        - php-fpm
```
wilson1000, line 40: I may have overwritten somewhere mate, this will cause an error, a loop, due to this line
https://github.com/ministryofjustice/justice-gov-uk/blob/main/docker-compose.yml#L21C9-L22C35

We also need port 80 listed in the local/server.conf for the integration above to work. 

This is all relevant unless something has changed 😄

deploy/config/local/server.conf
```diff
@@ -15,7 +15,6 @@ upstream php {
 
 server {
     listen  8080 default_server; # For default requests.
-    listen  80;                  # For localhost loopback only.
```
wilson1000, line 18: Just to note, this was added for: 
https://github.com/ministryofjustice/justice-gov-uk/blob/main/docker-compose.yml#L21C9-L22C35

docker-compose.yml
wilson1000: ... and this PR is probably that lol

public/app/themes/justice/inc/core.php
```diff
@@ -59,4 +62,42 @@ public function removeDefaultDashboardWidgets(): void
...
+     *
+     * Handle requests to the application host, by sending them to the loopback url.
+     *
+     * @param false|array|WP_Error $preempt
+     * @param array $args
+     * @param string $url
+     * @return false|array|WP_Error
+     */
+
+    public function handleLoopbackRequests(false|array|WP_Error $preempt, array $args, string $url): false|array|WP_Error
```
wilson1000, line 77: I found this slightly confusing buddy and had to look up the hook `pre_http_request`. 
The input variables are:

`$response`, `$parsed_args` and `$url`

Could we use the same names or indeed, make a note in the comment to map the meaning of the variable names?

public/app/themes/justice/inc/core.php
```diff
@@ -59,4 +62,42 @@ public function removeDefaultDashboardWidgets(): void
...
+            return $preempt;
+        }
+
+        // Is the request url to the application host?
+        if (parse_url($url, PHP_URL_HOST) !== parse_url(get_home_url(), PHP_URL_HOST)) {
+            return $preempt;
+        }
+
+        // Replace the URL.
+        $new_url = str_replace(get_home_url(), $loopback_url, $url);
```
wilson1000, line 92: Just a thought...
This will fire on every HTTP request called via `WP_Http::request()` 
Before modifying the URL, are we certain that every other type of request has been satisfied? For example; a call to a valid external site?

wilson1000 (dismissed):
This looks good @EarthlingDavey - there are some considerations around local functionality and external URLs but please, if these are not warranted, feel free to move ahead 😄

docker-compose.yml
EarthlingDavey: haha! yup!

public/app/themes/justice/inc/core.php
EarthlingDavey: Done!

public/app/themes/justice/inc/core.php
EarthlingDavey: Hey, good point!

This condition will catch if the request url host matches the host of WP_HOME

`if (parse_url($url, PHP_URL_HOST) !== parse_url(get_home_url(), PHP_URL_HOST))`

So, we're 100% only catching requests that are meant for the application.

wilson1000 (approved):
Nice!😊

PR #150, opened 2024-05-15
fixNationalArchiveLinks - fix the regex pattern that was not matching correctly.

EarthlingDavey:


wilson1000 (approved):
Looks good

public/app/themes/justice/inc/content.php
```diff
@@ -41,10 +41,10 @@ public function addHooks(): void
     public function fixNationalArchiveLinks($content)
     {
 
-        if (Config::get('WP_ENVIRONMENT_TYPE') !== 'production') {
-            // Match strings that start with https://webarchive.nationalarchives.gov.uk and contain: //dev //demo ///stage.
-            $pattern = '/(https:\/\/webarchive.nationalarchives.gov.uk.*)(\/\/dev)|(\/\/demo)|(\/\/stage)/';
-            return preg_replace($pattern, '$1//www', $content);
+        if (Config::get('WP_ENVIRONMENT_TYPE') === 'staging') {
+            // Match strings that start with https://webarchive.nationalarchives.gov.uk any amount words and / then ://stage
```
EmilyHazlehurst, line 45: @EarthlingDavey Not sure of the use case, but this regex won't pick up words with special characters e.g. `https://webarchive.nationalarchives.gov.uk/test-string://stage`

EarthlingDavey: Hey, thanks for the heads up! Good spot, but I think we're ok, it only needs to match strings like: `ukgwa/20220801160116`.

PR #151, opened 2024-05-15
Copy demo deployment to production

EarthlingDavey:
This doesn't use www.justice.gov.uk as ingress yet.

### DIFF
--- a/deploy/production/deployment.tpl.yml
+++ b/deploy/production/deployment.tpl.yml
@@ -1,0 +1,101 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ${KUBE_NAMESPACE}
+  namespace: ${KUBE_NAMESPACE}
+  labels:
+    app: ${KUBE_NAMESPACE}
+spec:
+  replicas: 2
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 0
+      maxSurge: 100%
+  selector:
+    matchLabels:
+      app: ${KUBE_NAMESPACE}
+  template:
+    metadata:
+      labels:
+        app: ${KUBE_NAMESPACE}
+    spec:
+      volumes:
+        - name: uploads
+          emptyDir: { }
+        - name: php-socket
+          emptyDir: { }
+      terminationGracePeriodSeconds: 35
+      serviceAccountName: ${KUBE_NAMESPACE}-service
+      containers:
+        - name: nginx
+          image: ${ECR_URL}:${IMAGE_TAG_NGINX}
+          ports:
+            - containerPort: 8080
+          volumeMounts:
+            - name: uploads
+              mountPath: /var/www/html/public/app/uploads
+            - name: php-socket
+              mountPath: /sock
+          resources:
+            requests:
+              cpu: 500m
+              memory: 200Mi
+            limits:
+              cpu: 500m
+              memory: 200Mi
+
+        - name: fpm
+          image: "${ECR_URL}:${IMAGE_TAG_FPM}"
+          ports:
+            - containerPort: 9000
+          volumeMounts:
+            - name: uploads
+              mountPath: /var/www/html/public/app/uploads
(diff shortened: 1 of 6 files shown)


deploy/production/deployment.tpl.yml
```diff
@@ -0,0 +1,101 @@
...
+              cpu: 500m
+              memory: 200Mi
+            limits:
+              cpu: 500m
+              memory: 200Mi
+
+        - name: fpm
+          image: "${ECR_URL}:${IMAGE_TAG_FPM}"
+          ports:
+            - containerPort: 9000
```
wilson1000, line 51: We don't need this anymore - sockets are king

deploy/production/ingress.yml
```diff
@@ -0,0 +1,53 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: justice-gov-uk-production-ingress
+  namespace: justice-gov-uk-production
+  annotations:
+    external-dns.alpha.kubernetes.io/set-identifier: justice-gov-uk-production-ingress-justice-gov-uk-production-green
+    external-dns.alpha.kubernetes.io/aws-weight: "100"
+    nginx.ingress.kubernetes.io/auth-type: basic
+    nginx.ingress.kubernetes.io/auth-secret: basic-auth-secret
```
wilson1000, line 10: We can remove basic auth, only we should reach it over host file redirect

deploy/production/ingress.yml
```diff
@@ -0,0 +1,53 @@
...
+    nginx.ingress.kubernetes.io/auth-realm: 'Production User | Authentication Required'
+    nginx.ingress.kubernetes.io/proxy-body-size: "200m"
```
wilson1000, line 12: Do we know what this means??

deploy/production/ingress.yml
```diff
@@ -0,0 +1,53 @@
...
+      location = /.well-known/security.txt {
+        auth_basic off;
+        return 301 https://raw.githubusercontent.com/ministryofjustice/security-guidance/main/contact/vulnerability-disclosure-security.txt;
+      }
+spec:
+  ingressClassName: default
+  tls:
+  - hosts:
+    - production-justice-gov-uk.apps.live.cloud-platform.service.justice.gov.uk
+  # - hosts:
```
wilson1000, line 30: We can update the local host file to reach this, is that what we discussed?

wilson1000 (commented):
Looks ok @EarthlingDavey - some comments I'd like to chat about first 😄

deploy/production/ingress.yml
EarthlingDavey: proxy-body-size is for allowing posing documents (zips) < 200 megs.

Ref: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#proxy-body-size

Looks like m (or M) is is ok to to use for megabytes https://nginx.org/en/docs/syntax.html

deploy/production/ingress.yml
EarthlingDavey: It was just a thought...

having `production-justice-gov-uk.apps.live.cloud-platform.service.justice.gov.uk` is one way to get the content up for QA.

Then prior to switchover, we change WP_HOME etc. and do a find/replace in the database.

deploy/production/ingress.yml
EarthlingDavey: Would be good to have a chat about this.

I don't think we factored in that we need to have Antony switch over the DNS, **then** get an ssl cert.

We could get around it...

- Create an a record to squiz. 
- Antony switches the DNS to Cloud Platform. 
- Squiz content is still being served.
- Add Certificate.yaml to cloud-platform-environments

deploy/production/deployment.tpl.yml
EarthlingDavey: Good point 🤣

wilson1000 (approved):

PR #152, opened 2024-05-15
Some minor deployment updates

wilson1000:
Manages basic auth
Fixes `make` error with wp-admin/index.php


docker-compose.yml
```diff
@@ -33,7 +33,6 @@ services:
         - ./deploy/config/local/server.conf:/etc/nginx/conf.d/default.conf
         - ./deploy/config/nginx.conf:/etc/nginx/nginx.conf
         - ./public/index.php:/var/www/html/public/index.php
```
EarthlingDavey, line 35: Will this still work locally from a fresh clone?

I think we don't have a script or a line in Dockerfile to get `wp-admin/index.php` into the nginx-dev container.

Makefile
```diff
@@ -1,10 +1,21 @@
 .DEFAULT_GOAL := d-shell
 
+##  █░█  ▄▀█  █▀█  █  ▄▀█  █▄▄  █░░  █▀▀  █▀
+##  ▀▄▀  █▀█  █▀▄  █  █▀█  █▄█  █▄▄  ██▄  ▄█
+##  populate as needed for testing
+##  ... never commit!
+COMPOSER_USER := ***
```
EarthlingDavey, line 7: Is it possible to grab these from .env?

.github/bin/deploy-run-script.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+BASIC_AUTH_BASE64=""
+
+## Prevent errors when basic auth isn't used
+##
+## Nb.the BASIC_AUTH_USER secret in GH production environment should
+## be set to `no-basic-auth` if not being used
+if [ "$BASIC_AUTH_USER" != "no-basic-auth" ]; then
```
EarthlingDavey, line 9: Really nice that this must be explicitly set.

I'd like to have a table in the readme at some point for the env vars. That would be a good place to make a note about how this works :)

EarthlingDavey (approved):

PR #155, opened 2024-05-16
Feature/cdpt 1626 review env vars for all environments add sentry plugin

EmilyHazlehurst:


wilson1000 (approved):
Looks good, can't wait to have Sentry operational on the frontend too 😄

PR #158, opened 2024-05-17
CDPT-1631: Review sitemap

EmilyHazlehurst:
- Add new sitemap class
- Remove the users sitemap from wp-sitemap.xml


EarthlingDavey (approved):
Looks great!

PR #162, opened 2024-05-17
Redirects for xhibit.

EarthlingDavey:
http://xhibit.justice.gov.uk/ is hot-linking assets from this site. This PR fixes 404s.


wilson1000 (approved):
🎉

PR #170, opened 2024-05-21
CDPT-1633: Post deploy infrastructure amends

EmilyHazlehurst:
- Prevent traffic from thousandeyes

I _think_ this should work, but I've not been able to test it locally so will need to check on dev.


wilson1000 (approved):
Let's give it a go 😄

PR #174, opened 2024-05-23
CDPT-1633: Move thousand eyes ingress block to prod

EmilyHazlehurst:
This is working on dev so we can move it to prod. Requests with the X-Thousandeyes-Agent header receive a 403 response.


wilson1000 (approved):
Looks good 😄

PR #178, opened 2024-05-28
CDPT-1665 Add html meta to match legacy site.

EarthlingDavey:
Some meta fields that were consistently empty on the legacy site have been removed.


wilson1000 (approved):
Looks good

PR #181, opened 2024-05-30
Sentry errors take two

wilson1000:
Removes unused code
Add safeguards for jQuery undefined
Adds save button
A good general clean up

### DIFF
--- a/public/app/themes/justice/src/patch/js/ccfw-gtm.js
+++ b/public/app/themes/justice/src/patch/js/ccfw-gtm.js
@@ -109,23 +109,15 @@ const CCFW = {
 
             if (remove) {
                 jQuery(element).attr('aria-checked', false)
-                jQuery('#ccfw-' + allowed + '-toggle-off').
-                removeAttr('aria-hidden').
-                show()
-                jQuery('#ccfw-' + allowed + '-toggle-on').
-                attr('aria-hidden', 'true').
-                hide()
+                jQuery('#ccfw-' + allowed + '-toggle-off').removeAttr('aria-hidden').show()
+                jQuery('#ccfw-' + allowed + '-toggle-on').attr('aria-hidden', 'true').hide()
             } else {
                 if (allowList.indexOf(allowed) === -1) {
                     allowList.push(allowed)
                 }
                 jQuery(element).attr('aria-checked', true)
-                jQuery('#ccfw-' + allowed + '-toggle-on').
-                removeAttr('aria-hidden').
-                show()
-                jQuery('#ccfw-' + allowed + '-toggle-off').
-                attr('aria-hidden', 'true').
-                hide()
+                jQuery('#ccfw-' + allowed + '-toggle-on'). removeAttr('aria-hidden'). show()
+                jQuery('#ccfw-' + allowed + '-toggle-off'). attr('aria-hidden', 'true'). hide()
             }
         })
 
@@ -208,33 +200,12 @@ const init = () => {
 
         // Always allow variables and triggers - https://developers.google.com/tag-manager/web/restrict
         let ccfwTriggers = [
-        'evl',
-        'cl',
-        'fsl',
-        'hl',
-        'jel',
-        'lcl',
-        'sdl',
-        'tl',
-        'ytl']
+            'evl', 'cl', 'fsl', 'hl', 'jel', 'lcl', 'sdl', 'tl', 'ytl'
+        ]
 
         let ccfwVariables = [
-        'k',
-        'v',
-        'c',
-        'ctv',
-        'e',
-        'jsm',
-        'dpg',
-        'd',
-        'vis',
-        'gas',
-        'f',
-        'j',
-        'smm',
-        'r',
-        'remm',
-        'u']
+            'k', 'v', 'c', 'ctv', 'e', 'jsm', 'dpg', 'd', 'vis', 'gas', 'f', 'j', 'smm', 'r', 'remm', 'u'
+        ]
 
         allowedList = allowedList.concat(ccfwTriggers, ccfwVariables)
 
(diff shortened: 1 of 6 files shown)


public/app/themes/justice/src/patch/js/ccfw-gtm.js
```diff
@@ -57,167 +69,189 @@ const CCFW = {
...
+                jQuery('#ccfw-all-toggle-on').removeAttr('aria-hidden').show()
+                jQuery('#ccfw-all-toggle-off').attr('aria-hidden', 'true').hide()
+                jQuery('button[data-allowlist="all"]').attr('aria-checked', true)
             }
         }
 
-        return allowList;
+        return allowList
+    },
+    hack: {
```
EarthlingDavey, line 154: Could we rename this it we're still thinking about pushing the work upstram?

public/app/themes/justice/src/patch/js/ccfw-gtm.js
```diff
@@ -57,167 +69,189 @@ const CCFW = {
...
+}
 
-const ccfwGTM = () => {
-    if (CCFW.canRun(CCFW.gtmID)) {
-        window.dataLayer = [];
+/**
+ * An init function to start the GTM feature
+ * @returns {boolean}
+ */
+const Init = () => {
```
EarthlingDavey, line 179: Should we stick with lower-case for function names?

public/app/themes/justice/src/patch/js/ccfw-gtm.js
```diff
@@ -1,22 +1,27 @@
...
             accept: 'cookie-accept',
             decline: 'cookie-decline'
         },
         moreInfo: 'cookie-more-info',
-        settings: 'js-ccfw-settings-button',
-        toggles: 'ccfw-banner__toggle-slider'
+        settings: 'js-ccfw-settings-button'
+    },
+    jq: {
+        toggles: jQuery('.ccfw-banner__toggle-slider'),
```
EarthlingDavey, line 21: If toggles is a jQuery object, what do you think about the convention of prefixing it with $? Ie. $toggles

public/app/themes/justice/src/patch/js/ccfw-gtm.js
```diff
@@ -57,167 +69,189 @@ const CCFW = {
...
 
         if (remove) {
-            allowList = [];
+            allowList = []
         }
 
-        jQuery('.' + CCFW.selector.toggles).each(function (key, element) {
-            let allowed = jQuery(element).data('allowlist');
+        CCFW.jq.toggles.each(function (key, element) {
+            let allowed = jQuery(element).data('allowlist')
```
EarthlingDavey, line 102: I'm not sure, but as toggles is a jQuery object, does element need to be wrapped in jQuery() here?

public/app/themes/justice/src/patch/js/ccfw-gtm.js
```diff
@@ -57,167 +69,189 @@ const CCFW = {
...
 
-const togglesChange = function (e) {
-    e.preventDefault();
-    let toggle = jQuery(this);
-    let allowed = toggle.data('allowlist');
-    let allowList = CCFW.storage.allowed.get() || [];
-    let pressed = toggle.attr('aria-checked') === 'true';
-    toggle.attr('aria-checked', !pressed);
+    e.preventDefault()
+    let toggle = jQuery(this)
```
EarthlingDavey, line 228: If we stick with semi-colons at line endings, it will be easier for upstream to review and merge.

EarthlingDavey (changes requested):

public/app/themes/justice/src/patch/js/ccfw-gtm.js
wilson1000: Sure 👍🏼

public/app/themes/justice/src/patch/js/ccfw-gtm.js
wilson1000: I have experienced the convention confusing for new developers. Some think you can only cache a jQuery object if you place a $ at the beginning. This isn't true, JavaScript always passes objects by reference, in effect; creating cache. I would much rather code as the language is intended without the added confusion. This is why I place the object within jq

public/app/themes/justice/src/patch/js/ccfw-gtm.js
wilson1000: Element is actually a DOM object ☺️ we convert it here

wilson1000: https://api.jquery.com/Types/#Element

public/app/themes/justice/src/patch/js/ccfw-gtm.js
wilson1000: Thanks for pointing this out @EarthlingDavey 
My IDE strips them and then sometimes doesn't. It's weird behaviour.
When laravel mix does its work I think it adds them all anyway.
I'll try and fix it locally. Cheers

wilson1000: Hey @EarthlingDavey I checked the [JavaScript standard and it states no semicolons](https://standardjs.com/rules#semicolons). Off the back of that, I've removed all semicolons in our patch code.

I'm more than happy to talk about this and hear your point of view if you'd like to.

public/app/themes/justice/src/patch/js/ccfw-gtm.js
wilson1000: For reference:
https://api.jquery.com/each/#each-function

```javascript
$( "div" ).each(function( index, element ) {
    // element == this ... a document object model DIV element
    $( element ).css( "backgroundColor", "yellow" );
    if ( $( this ).is( "#stop" ) ) {
      $( "span" ).text( "Stopped at div index #" + index );
      return false;
    }
  });
```

EarthlingDavey: Ah, I see, thanks for explaining :D

EarthlingDavey (approved):

EarthlingDavey (approved):

PR #183, opened 2024-06-04
Error handling for cookies disabled

wilson1000:
When local storage or cookies are disabled, the application reports this back without throwing an error.

The report is simply that GTM couldn't be initiated.

Screenshot 2024-06-04 at 19 03 28

Includes a cheeky update in Dockerfile... adds `screenshot.png` to the Nginx container file system.


public/app/themes/justice/src/patch/js/ccfw-storage.js
```diff
@@ -1,29 +1,82 @@
...
+    },
+    cookie: () => {
+        try {
+            // try and set it...
+            document.cookie = 'cookie_test=1'
+            // test it exists...
+            const cookiesEnabled = document.cookie.indexOf('cookie_test=') !== -1
+            // remove it...
+            document.cookie = 'cookie_test=1; expires=Thu, 01-Jan-1970 00:00:01 GMT'
+            return cookiesEnabled
```
wilson1000, line 43: This is an odd one...

> ... in some in situations `document.cookie` is exposed but throws a
> SecurityError if you try to access it...

https://github.com/Modernizr/Modernizr/blob/master/feature-detects/storage/cookies.js

We checked the try/catch result in case no error was thrown but we still couldn't write the cookie.

public/app/themes/justice/src/patch/js/ccfw-storage.js
```diff
@@ -1,29 +1,82 @@
...
+    },
+    local: () => {
+        let mod = 'ccfw';
+        try {
+            localStorage.setItem(mod, mod);
+            localStorage.removeItem(mod);
+            return true;
+        } catch (e) {
+            return false
+        }
```
wilson1000, line 33: https://github.com/Modernizr/Modernizr/blob/master/feature-detects/storage/localstorage.js

public/app/themes/justice/src/patch/js/ccfw-storage.js
```diff
@@ -1,29 +1,82 @@
...
+ * we disable storage all together.
  */
-if (window.localStorage === undefined) {
-    Storage = {
-        getItem: (key) => {
-            const value = `; ${document.cookie}`;
-            const parts = value.split(`; ${key}=`);
+let Storage = {
+    disabled: true
+}
```
wilson1000, line 9: By default, we start with a disabled Storage object. We can test directly against this.
The logic is that this object is overwritten if localStorage or Cookies are available.

public/app/themes/justice/src/patch/js/ccfw-storage.js
EarthlingDavey: Great stuff finding references to Modernizr functions 🚀

public/app/themes/justice/src/patch/js/ccfw-storage.js
```diff
@@ -1,29 +1,82 @@
...
+        try {
+            // try and set it...
+            document.cookie = 'cookie_test=1'
+            // test it exists...
+            const cookiesEnabled = document.cookie.indexOf('cookie_test=') !== -1
+            // remove it...
+            document.cookie = 'cookie_test=1; expires=Thu, 01-Jan-1970 00:00:01 GMT'
+            return cookiesEnabled
+        } catch (e) {
+            // Catch and handle the error if cookies are disabled.
```
EarthlingDavey, line 45: Possibly rename to `Catch and ignore`?

public/app/themes/justice/src/patch/js/ccfw-storage.js
```diff
@@ -1,29 +1,82 @@
...
+            return false
+        }
+    }
+}
 
-            if (parts.length === 2) {
-                let part = parts.pop().split(';').shift();
-                return part === '' ? null : part;
+if (CCFWStorage.valid()) {
```
EarthlingDavey, line 51: A minor, but could we avoid nested `if`s here?

public/app/themes/justice/src/patch/js/ccfw-gtm.js
```diff
@@ -188,6 +188,14 @@ const CCFW = {
  * @returns {boolean}
  */
 const init = () => {
+    /**
+     * We cannot run if we don't have a storage object
+     * Run a test to check the Storage engine
+     */
+    if (Object.hasOwn(Storage, 'disabled')) {
+        return false;
+    }
```
EarthlingDavey, line 197: Another minor, I suppose this could potentially introduce a bug further down the line, where another developer may stick `disabled: false` on the Storage object - this conditional would be true.

It would be a bit safer to explicitly check if disabled is true here.

EarthlingDavey (commented):
Nice! Appreciate the code & GH comments! Just a few queries.

public/app/themes/justice/src/patch/js/ccfw-storage.js
```diff
@@ -1,29 +1,82 @@
...
+     * If cookie returns true, valid() returns true, otherwise, valid() returns
+     * false.
+     *
+     * @returns {boolean}
+     */
+    valid: () => {
+        return CCFWStorage.local() || CCFWStorage.cookie()
+    },
+    local: () => {
+        let mod = 'ccfw';
```
EarthlingDavey, line 26: I think this should be assigned with `const` vs `let`.

what do you think about moving this declaration to the top of the file, then also use it for the cookie test, to save a few bytes.

e.g. 

```js
document.cookie = `${mod}=1`
```

public/app/themes/justice/src/patch/js/ccfw-storage.js
wilson1000: I see your point and value it, however, the `valid()` call here must encompass checks for `local()` and `cookie()`. When we pass valid(), we know that either local() or cookie() is alive, we have to work out which one. If `local()` is not accessible, we know that `cookie()` is.

EarthlingDavey:
Last comment! ... It may be good to version bump in `public/app/themes/justice/inc/plugin-hacks.php`, to clear any browser cache.

public/app/themes/justice/src/patch/js/ccfw-storage.js
wilson1000: Cool point :) although, in this case, I'd like to keep the checks independent. Using a const for both makes sense. Thank you.

public/app/themes/justice/src/patch/js/ccfw-gtm.js
wilson1000: It's not the best option but definitely progress. `disabled` won't exist on the Storage object in most cases. I've added a note above the object declaration to help in this circumstance. Thanks for pointing it out.

EarthlingDavey (approved):

PR #192, opened 2024-06-11
Feature/cdpt 1704 UI for inline anchor links

EarthlingDavey:
This PR adds support for anchor destinations to rich-text content in the block editor.

It does the following:
- adds an anchor button to the toolbar.
- allows the user to enter and clear an anchor id.
- formats legacy anchor links that would otherwise be invisible to editors.
- has user's preference for showing anchor icons.


wilson1000 (approved):
Hey @EarthlingDavey - great work here mate. Thanks.

PR #195, opened 2024-06-12
Feature/cdpt 1704 - Amends - UI for inline anchor links

EarthlingDavey:
- Amend validation to allow existing ids , and use named attributes
- Block editor page css fix. Open page button hidden was hidden, white space to the left of the update button. (@wilson1000 - a minor fix that I've bundled in here by increasing the specificty).

image


wilson1000 (approved):
Let's roll with this 😄

PR #196, opened 2024-06-12
feature/cdpt 1534 Use nginx for redirects

EarthlingDavey:


EarthlingDavey:
Thanks @wilson1000 ... yes this should be more efficient. Rather than nginx -> php -> db (and back again).

The ability for the editors to add redirects was a requested feature. I propose we keep the plugin, and keep an eye on it's reliability with a handful of redirects - what do you think?

wilson1000 (approved):
Looks good, thanks mate

PR #197, opened 2024-06-13
Fix error:  Cannot read properties of null (reading 'classList')

EarthlingDavey:
Uses body as the element to toggle a user preference variation class. Instead of `.wp-block-post-content`, which is not in the DOM for the code editor view.

Fixes: https://ministryofjustice.sentry.io/issues/5481194459/?environment=production&project=4506813195157504&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=24h&stream_index=4


wilson1000 (approved):
Nice 👍

PR #198, opened 2024-06-13
Update wp-document-revisions

EarthlingDavey:
- Addresses an archived issue in sentry `window.WPDocumentRevisions is undefined`
- Removes hotfix from composer-post-install. As it's been released into 3.6.1


wilson1000 (approved):
Looks good 😄

PR #203, opened 2024-06-13
Feature/cdpt 1534 use nginx for redirects amends

EarthlingDavey:
nginx is running in the context that incoming requests are http and port 8080.

This can be seen by a bug that I've identified e.g. on stage https://stage.justice.gov.uk/wp-content/uploads/test will redirect to http://stage.justice.gov.uk:8080/app/uploads/test , then https://stage.justice.gov.uk:8080/app/uploads/test

This fix will use relative paths for redirects, avoiding the above issue.

See: https://www.rfc-editor.org/rfc/rfc9110.html#section-10.2.2

### DIFF
--- a/deploy/config/local/server.conf
+++ b/deploy/config/local/server.conf
@@ -25,6 +25,9 @@ server {
     root /var/www/html/public;
     index index.php index.html;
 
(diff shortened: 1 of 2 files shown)


wilson1000 (approved):
Hey mate, thanks for the info.

PR #210, opened 2024-06-19
CDPT-1740: Search panel not working

EmilyHazlehurst:
- Clean up parent filter function
- Add hidden parent field to panel search


EmilyHazlehurst:
The ticket said _This panel could be renamed to 'Standard direction search'._

I haven't done this because the header has a fixed height, that title takes it over two lines and the CSS is difficult to change in it's current state. 

I've made a note on the appropriate frontend ticket to change it.

public/app/themes/justice/inc/search.php
```diff
@@ -265,20 +265,22 @@ public function relevanssiParentFilter(array $hits): array
...
             return $hits;
+        } else {
+            $parent_id = $wp_query->query_vars['parent'];
         }
 
         $acc = [];
         foreach ($hits[0] as $hit) {
             // Loop through all the posts found.
-            if ($hit->ID === $wp_query->query_vars['parent']) {
+            if ($hit->ID === $parent_id) {
```
wilson1000, line 275: That looks like a swear word 😄

public/app/themes/justice/inc/search.php
```diff
@@ -265,20 +265,22 @@ public function relevanssiParentFilter(array $hits): array
         if (!isset($wp_query->query_vars['parent'])) {
             // No parent parameter set, do nothing.
             return $hits;
+        } else {
```
wilson1000, line 268: You don't need else here, the if will return, meaning no dice.

```php
if (!isset($wp_query->query_vars['parent'])) {
    // No parent parameter set, do nothing.
    return $hits;
}

$parent_id = $wp_query->query_vars['parent'];
```

A general guide I like to follow in coding: **_if else, why?_**

public/app/themes/justice/inc/search.php
wilson1000, line 275: `switch` is faster than multiple if, else

```php
switch (true) {
    case ($hit->ID === $parent_id):
        break;
    case ($hit->post_parent === $parent_id): 
        break;
    case ($hit->post_parent > 0):
        break;
}
```

wilson1000 (approved):
Looks good! I've left some little comments, nothing too serious and approved as this is fixing the issue with some amends - progression always wins 👍

public/app/themes/justice/inc/search.php
EmilyHazlehurst: Thanks @wilson1000, I didn't actually write this it was already there. I think the formatter just adjusted a few things, happy to refine it though :) 

I'm not sure that's always true, PHP Bench reckons they're about the same https://phpbench.com/
Screenshot 2024-06-20 at 09 33 41

In which case I suppose it's down to our coding standards/readability.

wilson1000: Yes, I'd challenge that. I did a benchmark a while ago and the difference was quite a bit. Although, there's nothing to say the speed of `multi if else` hasn't improved with age 😄 - thanks for sharing

wilson1000: So here's a funny thing... I clicked the phpbench link and my browser froze lol

wilson1000 (approved):
Looks great, thanks @EmilyHazlehurst

public/app/themes/justice/inc/search.php
EmilyHazlehurst: Haha, maybe PHP Bench isn't the most reliable then... 😆

PR #217, opened 2024-07-02
Feature/cdpt 1714 container setup

EmilyHazlehurst:


public/app/frontend/style.scss
```diff
@@ -0,0 +1,312 @@
...
+    'sm': 1.5rem, // The smallest icon must be at least 24px x 24px to meet WCAG 2.2 AA
+    'md': 3rem,
+    'lg': 6rem
+);
+
+.icon {
+    @each $size, $rem in $iconSizes {
+        @each $icon, $ext in $iconTypes {
+            &-#{$icon}--#{$size} {
+                background-image: url(/src/assets/icons/#{$icon}.#{$ext});
```
EarthlingDavey, line 253: Is there a method to decide if a component should have global styles here, like .icon ... or be a mixin like `@mixin link` below?

This is something where I've not really known for sure on previous projects.

public/app/frontend/src/layouts/with-sidebar/with-sidebar.stories.js
```diff
@@ -0,0 +1,23 @@
...
+
+export const Right = Template.bind({});
+Right.args = {
+    direction: 'right',
+};
+
+export const Left = Template.bind({});
+Left.args = {
+    direction: 'left',
+};
```
EarthlingDavey, line 23: Should we have a story for 2 sidebars here?

public/app/frontend/style.scss
```diff
@@ -0,0 +1,312 @@
...
+    display: grid;
+    grid-template-columns: repeat(4, [col-start] 1fr);
+    gap: $gap-mobile;
+    width: 100%;
+    @media screen and (min-width: $justice-lg) {
+        grid-template-columns: repeat(12, [col-start] 1fr);
+        gap: $gap;
+    }
+}
+
```
EarthlingDavey, line 101: Looks good :)

EarthlingDavey (approved):
Loos great :)

public/app/frontend/src/layouts/with-sidebar/with-sidebar.stories.js
EarthlingDavey: nvm, found it.

public/app/frontend/style.scss
EmilyHazlehurst: @EarthlingDavey good question.

I think it this case it was purely so that I could use it in the Storybook icon examples page without having to style the page, but that's not a very good reason so I'll change it 😅 

Generally I would prefer most things to be a mixin rather than a global style. Particularly with BEM it means we're namespacing everything.

In this case if we used `.icon` and Wordpress used `.icon` there could be some conflicts, whereas if we have `header__icon` with `@include icon` we're less likely to have that issue.

EarthlingDavey: Nice, that makes sense!

PR #219, opened 2024-07-04
Feature/cdpt 1745 component link

EmilyHazlehurst:


EarthlingDavey (approved):

public/app/frontend/style.scss
```diff
@@ -310,3 +311,43 @@ $iconSizes: (
...
+
+    // For long content, line feeds are not interpreted as spaces and small width
+    // causes content to wrap 1 word per line:
+    // https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe
+    white-space: nowrap if($important, !important, null);
+
+    // Prevent users from selecting or copying visually-hidden text. This prevents
+    // a user unintentionally copying more text than they intended and needing to
+    // manually trim it down again.
+    user-select: none;
```
EarthlingDavey, line 352: As some of this is copied from GDS, would it be a bad idea to install the GDS (or even MoJ) design system npm package and import from it?

EarthlingDavey (approved):

public/app/frontend/style.scss
EmilyHazlehurst: @EarthlingDavey I've tried to keep it separate as much as possible because the justice site has it's own visual identity and it's only a couple of things that we're using from the design system. 

I'm hoping in future we can use the GDS or MoJ design system but given the concerns around any major visual changes I think it's best to try to keep them apart in these initial stages, though I'll revisit it if we find we're using a lot from them.

EarthlingDavey: That sounds good. Would be nice to see the site head in the direction of GDS or MoJ _at some point_.

PR #221, opened 2024-07-04
Feature/cdpt 1747 component nav main

EmilyHazlehurst:


public/app/frontend/src/components/navigation-main/navigation-main.html.twig
```diff
@@ -0,0 +1,44 @@
+{#
+@file
+The main navigation for the site
+
+Available variables:
+    - links: array An array of links to be displayed
+        - url: string The url for the link
+        - label: string The label for the link
```
EarthlingDavey, line 8: I think this is missing newTab.

public/app/frontend/base.scss
```diff
@@ -40,6 +40,13 @@ a, .link {
     @include link;
 }
 
+ul {
+    list-style: none;
+    margin:0;
+    padding: 0;
+    text-indent: 0;
+}
+
```
EarthlingDavey, line 49: If we set this globally - we might end up having to override the reset when we get to styling 'Rich Text' page content.

I'm don't have strong views either way, but something to consider, if you haven't already :)

EarthlingDavey (approved):

public/app/frontend/base.scss
EmilyHazlehurst: @EarthlingDavey that's a fair point. I tend to prefer resetting ol as it's so often used in nav structures and in the rich text PR I've set up the correct styling there.

PR #222, opened 2024-07-08
Feature/cdpt 1751 search bar

EmilyHazlehurst:


public/app/frontend/src/components/search-bar/search-bar.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    {% include '@components/search-bar/search-bar.html.twig' with {
+        action: '/',
+        id: 'search-top'
+        input: {
+            labelHidden: true,
+            label: 'Search',
+            id: 'search-top-input',
+        },
+        button: {
+            style: 'primary',
```
EarthlingDavey, line 20: Super minor, as style is sometimes used in jsx as css. e.g. `style={{'padding':'10px'}}`.

Could we use `variant: 'primary'` or something else other than style?

public/app/frontend/src/components/navigation-main/navigation-main.scss
```diff
@@ -13,6 +13,7 @@
         width: 100%;
         @media screen and (min-width: $justice-lg) {
             flex-direction: row;
+            margin-left: -0.625rem;
```
EarthlingDavey, line 16: Could this be replaced with a new variable, or a a fraction of an existing variable like `$gutter / 2`?

EarthlingDavey (approved):

public/app/frontend/src/components/search-bar/search-bar.html.twig
EmilyHazlehurst: Good shout! I've updated that :)

public/app/frontend/src/components/navigation-main/navigation-main.scss
EmilyHazlehurst: Done :)

PR #224, opened 2024-07-09
Feature/cdpt 1731 footer

EmilyHazlehurst:


EarthlingDavey (approved):
Looks good, this might be a good PR to make the footer hit the bottom of the page on pages that aren't full screen height.

I've checked the layouts, I don't _think_ it's done already.

PR #227, opened 2024-07-10
Build and watch node scripts in parallel and in node container

EarthlingDavey:
This PR adds the ability to run the frontend and theme node scripts in parallel with docker.

Pros:

- we don't need to run individual commands for multiple services.
- devs don't need to run node directly on their host machine.

Cons:

- the terminal output isn't pretty, maybe this could be improved by using 3 node containers, one for each script.

Would be good to know your thoughts @EmilyHazlehurst


EmilyHazlehurst (approved):

docker-compose.yml
```diff
@@ -50,19 +52,27 @@ services:
...
+      # Theme
+      - theme_node_modules:/node/theme/node_modules
+      - theme_dist:/node/theme/dist
+      - ./public/app/themes/justice/src:/node/theme/src
+      - ./public/app/themes/justice/jsconfig.json:/node/theme/jsconfig.json
+      - ./public/app/themes/justice/package.json:/node/theme/package.json
+      - ./public/app/themes/justice/package-lock.json:/node/theme/package-lock.json
+      - ./public/app/themes/justice/webpack.mix.js:/node/theme/webpack.mix.js
+      # Frontend
+      - ./public/app/frontend:/node/frontend
```
EmilyHazlehurst, line 70: @EarthlingDavey running this locally doesn't make the frontend available to the theme. Do we need to add       
```- frontend_node_modules:/node/frontend/node_modules
- frontend_dist:/node/frontend/dist
- ./public/app/frontend/src:/node/frontend/src
- ./public/app/frontend/package.json:/node/frontend/package.json
- ./public/app/frontend/webpack.config.js:/node/frontend/webpack.config.js
```      
as we have with the theme?

EmilyHazlehurst: I've added `main.css` and `main.js` to the loadStyles and loadScripts functions in `theme-assets.php` but get a "not found" error in the console.

EarthlingDavey: I'll take a look... with the theme we were selective because there are a load of php files that didn't need to be available in the node container.

Opposite to this, pretty much every file in the frontend folder needs to be available to the node container, that's what line 70 is doing - we don't need specify individual files.

I've missed a couple of things (that I can add)

- Named volumes for the frontend's `dist` and `node_modules` having them named instead of mounted to the client means faster read/write.
- Sharing `dist` with the nginx container, do that those files are available to be served.

PR #229, opened 2024-07-17
Feature/cdpt 1744 image with text

EmilyHazlehurst:

PR #230, opened 2024-07-17
Feature/cdpt 1739 file download

EmilyHazlehurst:


EmilyHazlehurst:
Based on https://github.com/ministryofjustice/justice-gov-uk/pull/229

public/app/frontend/src/components/navigation-main/navigation-main.html.twig
```diff
@@ -26,7 +26,7 @@ example:
         ]
     }%}
 #}
-<nav id="navigation-main" class="navigation-main">
+<nav id="navigation-main" class="navigation-main" aria-label="Primary navigation">
```
EmilyHazlehurst, line 29: This should be aria-label="Primary" otherwise the screen reader will say "Primary navigation, navigation"

EarthlingDavey (commented):
This looks good on it's own but I think it would be helpful to add some context.

[This page](https://www.justice.gov.uk/courts/procedure-rules/family/update) is an example of where an editor has created a table in the block editor, and created links to zips.

And, I think there are other pages where editors have added these download links inline with paragraphs.

At the moment, the mechanism for adding one of the icons is to create a link in the editor, and set the href to a document. Based on the file extension, a download icon is shown.

To rollout this block should we ask the editor to continue adding documents in the same way? And use some backend code to filter content:

e.g. from 
```
<p>some text <a href="https://www.justice.gov.uk/courts/procedure-rules/family/zips/fpr-update-5-html.zip">HTML</a> </p>
```

to

```
<p>some text <div class="file-download">
    <i class="file-download__icon icon-{{ format|lower }}--sm" aria-hidden="true"></i>
    <a class="file-download__link" href={{ link }}>
        <span class="file-download__prefix visually-hidden">Download</span>
        {{ filename }}
    </a>
    <span class="file-download__details">({{ language ? language ~ ", " : "" }}{{ format }}, {{ filesize }})</span>
</div></p>
```

EmilyHazlehurst:
> This looks good on it's own but I think it would be helpful to add some context.
> 
> [This page](https://www.justice.gov.uk/courts/procedure-rules/family/update) is an example of where an editor has created a table in the block editor, and created links to zips.
> 
> And, I think there are other pages where editors have added these download links inline with paragraphs.
> 
> At the moment, the mechanism for adding one of the icons is to create a link in the editor, and set the href to a document. Based on the file extension, a download icon is shown.
> 
> To rollout this block should we ask the editor to continue adding documents in the same way? And use some backend code to filter content:
> 
> e.g. from
> 
> ```
> <p>some text <a href="https://www.justice.gov.uk/courts/procedure-rules/family/zips/fpr-update-5-html.zip">HTML</a> </p>
> ```
> 
> to
> 
> ```
> <p>some text <div class="file-download">
>     <i class="file-download__icon icon-{{ format|lower }}--sm" aria-hidden="true"></i>
>     <a class="file-download__link" href={{ link }}>
>         <span class="file-download__prefix visually-hidden">Download</span>
>         {{ filename }}
>     </a>
>     <span class="file-download__details">({{ language ? language ~ ", " : "" }}{{ format }}, {{ filesize }})</span>
> </div></p>
> ```

Thanks @EarthlingDavey, yes I was planning on doing something like that when integrating it into the backend. 

I've added the code into this PR for completeness, but essentially I'm using the `render_block` hook to replace the existing link with the new template. 

My main concern is that the editors have accounted for things like the filesize not being calculated automatically, so when we replace the link it appears twice in some places:

Screenshot 2024-07-29 at 11 05 04

Depending on how many instances of this type of link there are we could potentially update them manually. 

Ultimately I don't want the editors to have to make any changes to how they add content so it may be that we don't use this in the end, but it's a good template for other block components so probably worth keeping for now.

public/app/frontend/src/components/file-download/file-download.stories.js
```diff
@@ -0,0 +1,33 @@
...
+const Template = (args) => {
+    return fileDownload(args);
+};
+
+export const Default = Template.bind({});
+Default.args = {
+    format: 'PDF',
+    link: '#',
+    filesize: '167 KB',
+    filename: 'CPR email address list',
```
wilson1000, line 26: If these are default, should we be specifying initial values that suggest they are not currently set:
```
    filesize: '0 KB',
    filename: '',
```

public/app/frontend/src/components/navigation-main/navigation-main.html.twig
```diff
@@ -26,7 +26,7 @@ example:
         ]
     }%}
 #}
-<nav id="navigation-main" class="navigation-main">
+<nav id="navigation-main" class="navigation-main" aria-label="Primary">
```
wilson1000, line 29: Does the main navigation need Primary as an aria label, and will the end user that requires this label understand the context?

public/app/themes/justice/functions.php
```diff
@@ -96,6 +97,9 @@
 $taxonomies = new Justice\Taxonomies();
 $taxonomies->addHooks();
 
+$templates = new Justice\Templates();
+$templates->addHooks();
```
wilson1000, line 101: This will create `$templates` in the global scope. Are we sure nothing will overwrite this later, during PHP execution?

wilson1000 (approved):
Hi Emily, this all looks great - I've added some comments, one about aria is dumb but wanted to double-check with you.

public/app/frontend/src/components/file-download/file-download.stories.js
EmilyHazlehurst: Hi @wilson1000, they're not default values, anything in the .stories.js files are example values that are used when displaying the story. They won't appear outside of storybook.

public/app/frontend/src/components/navigation-main/navigation-main.html.twig
EmilyHazlehurst: @wilson1000 Good question. In this case we need an aria label because the justice site has multiple nav bars, this one and the sidebar. Whether Primary and Secondary are enough to explain their purpose I'm not sure, but it does at least differentiate between them. 

The MOJ design system example (https://design-patterns.service.justice.gov.uk/components/primary-navigation/) uses primary.

I think ideally we would have a "mega menu" as the main menu which integrates the dynamic parts of the sidebar, but as we can't change the page layout we're a bit limited unfortunately.

PR #233, opened 2024-08-02
CDPT-1755: To the top

EmilyHazlehurst:
- Add to the top template
- Replace gutenberg block in the backend


public/app/themes/justice/inc/templates.php
```diff
@@ -61,6 +62,7 @@ public function replaceWordpressBlocks(string $block_content, array $block): str
             switch ($block['blockName']) {
                 case 'core/paragraph':
                 case 'core/table':
+                case 'moj/to-the-top':
```
EarthlingDavey, line 65: I'm thinking this loop might get a bit slow for pages with 100s of blocks.

WordPress by default doesn't loop over blocks in this way - usually what is in the database is very close to the final rendered output. 

As `moj/to-the-top` will always have the same content do we need to run it through `renderLinks`?

In the case of `moj/to-the-top` could we simplify things by doing the equivalent of `$block_content = <use the to-the-top template with these static properties>`?

EarthlingDavey (approved):
Hey @EmilyHazlehurst , what you've done looks good, but I have some doubts about weather it could be simplified.

We have quite a few files and some DOM creating and parsing, for a what is (or nearly is nearly?) a static component.

Do you think it would be possible to simplify it without losing the benefits of the standalone frontend folder & Storybook?

EmilyHazlehurst:
> Hey @EmilyHazlehurst , what you've done looks good, but I have some doubts about weather it could be simplified.
> 
> We have quite a few files and some DOM creating and parsing, for a what is (or nearly is nearly?) a static component.
> 
> Do you think it would be possible to simplify it without losing the benefits of the standalone frontend folder & Storybook?

@EarthlingDavey thanks for the feedback. 

I understand what you're saying, my thoughts were that blocks should be handled consistently so that the codebase is easy to read and understand. 

I did a load speed comparison between the original component and this one and there wasn't a noticeable difference, but you're right in that it might be a problem as more blocks are made or on longer pages.

I'm going to merge this now but have made a ticket to look into this again after the static components are finished just to ensure it won't add unnecessary overheads and to look at other options.

PR #237, opened 2024-08-06
Feature/cdpt 1748 navigation sidebar

EmilyHazlehurst:


public/app/themes/justice/page_home.php
```diff
@@ -1,41 +1,22 @@
...
-    </div>
-    <div id="footer-bar">
-        <div>
-        </div>
-    </div>
-<?php
-
 get_footer();
+
+$templates = array( 'templates/home.html.twig');
```
wilson1000, line 10: Hi Emily, I'm loving the work!
One observation: throughout the file, we can see [] arrays, can we use those here?

wilson1000 (approved):
Looks good 🎉

PR #239, opened 2024-08-16
Update ingress class to modsec

EarthlingDavey:
Previous production ingress has been retained and set to weight 0, incase we need to rollback.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ WORKDIR /var/www/html
 
 ###
 
-FROM nginx:1.26-alpine as nginx-module-builder
+FROM nginx:1.26.1-alpine AS nginx-module-builder
 
 SHELL ["/bin/ash", "-exo", "pipefail", "-c"]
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -21,7 +21,7 @@ WORKDIR /var/www/html
 
 ###
 
-FROM nginx:1.26-alpine as nginx-module-builder
+FROM nginx:1.26.1-alpine AS nginx-module-builder
```
EarthlingDavey, line 24: This hass been pinned to address version drift between `nginx` and `nginxinc/nginx-unprivileged`

deploy/production/ingress.yml
```diff
@@ -40,3 +40,59 @@ spec:
...
+metadata:
+  name: justice-gov-uk-production-ingress-modsec
+  namespace: justice-gov-uk-production
+  annotations:
+    external-dns.alpha.kubernetes.io/set-identifier: justice-gov-uk-production-ingress-modsec-justice-gov-uk-production-green
+    external-dns.alpha.kubernetes.io/aws-weight: "100"
+    nginx.ingress.kubernetes.io/enable-modsecurity: "true"
+    nginx.ingress.kubernetes.io/modsecurity-snippet: |
+      # SecAuditEngine On
+      # SecRuleEngine DetectionOnly
```
EmilyHazlehurst, line 55: @EarthlingDavey are these supposed to be commented out/can they be removed if we're not using?

EarthlingDavey: Good shout, I've removed them.

EmilyHazlehurst (approved):

PR #241, opened 2024-08-22
Styling tweaks

EmilyHazlehurst:
- Add sidebar decorator so sidebar elements are displayed in a more representative fashion
- Set UTF charset in base scss file so >> element is displayed consistently in breadcrumb element


wilson1000 (approved):
Looks good 👍

PR #242, opened 2024-08-22
Accessibility improvements and link updates

EmilyHazlehurst:
- Add correct links to footer and header in WP
- Add visually hidden prop for links that should open in new tabs - this lets us keep the structure of the header/footer whilst still communicating page changes to screen reader users
- Hide first gov.uk link from footer so that it's not repeated for screen reader users


wilson1000 (approved):
Looks good 👍

PR #245, opened 2024-08-22
Sidebar adjustments

EmilyHazlehurst:
- Use the same structure for the single sidebar layout as is used in the two sidebar one so mobile panels are correctly positioned.
- Fix 'no descending selector specificity' error in scss


wilson1000 (approved):
Nice!

PR #260, opened 2024-09-11
Feature/cdpt 1855 create a build step to deploy storybook

EmilyHazlehurst:


.github/workflows/design-system.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: 'Build Node'
+        shell: bash
+        run: npm run build-storybook
+
+      - name: "Configuring AWS credentials"
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.S3_ROLE_TO_ASSUME }}
+          aws-region: "eu-west-2"
+
```
EmilyHazlehurst, line 44: I opted not to use the storybook-s3 action we discussed as it was just using configure-aws-credentials underneath so this is a bit cleaner.

EmilyHazlehurst:
Hi @EarthlingDavey, I think this should work now provided all of the permissions are correct on Cloud Platform. I've simplified it so we're just building and copying the files in Github Actions, but let me know if anything looks wrong. Thanks :)

.github/workflows/design-system.yml
EarthlingDavey: Sounds good!

EarthlingDavey (approved):

PR #274, opened 2024-10-10
Feature/cdpt 2075 search bar errors

EmilyHazlehurst:


public/app/themes/justice/search.php
```diff
@@ -71,14 +71,14 @@
...
 
     // If the post type is document get the format and filesize
     if ($document->verify_post_type($postId)) {
         $upload_dir = $document->document_upload_dir();
         $year_month = str_replace('-', '/', substr($result->post_date, 0, 7));
         $format = $document->get_file_type($postId);
-        $document_url = "{$upload_dir}/{$year_month}/{$result->post_title}{$format}";
+        $document_url = "{$upload_dir}/{$year_month}/{$result->slug}{$format}";
         if (file_exists($document_url)) {
             $filesize = size_format(wp_filesize($document_url));
```
EarthlingDavey, line 83: While this code wasn't introduce in this PR, I've spotted something that might be a bug.

As we are offloading documents to S3, they won't be on the filesystem - I think checking the file size like this won't work.

However, WP Offload Media does save the file size in the database, there is a function that you can either use directly or copy. This screenshot should give an overview of the function and database structure.

image

public/app/themes/justice/search.php
```diff
@@ -98,11 +98,29 @@
...
 $pagination = $results->pagination();
+$didYouMean = [];
 $suggestedTerm = relevanssi_premium_generate_suggestion($query);
-$didYouMean = [
-    'term' => $suggestedTerm,
-    'url' => `search/${suggestedTerm}`
-];
+// If the term used is valid relevanssi_premium_generate_suggestion returns true
+// We don't need to display anything if this is the case
+if ($suggestedTerm && $suggestedTerm !== true) {
```
EarthlingDavey, line 105: `if ($suggestedTerm !== true) {` might be ok here.

public/app/themes/justice/search.php
EarthlingDavey, line 1: `get_footer();` should be at the end of the page. For example some script tags are added by `get_footer` and they are expected to be just before the closing `body` tag.

EarthlingDavey (approved):
Nice work. I've run it on localhost and it's working well!

I've written some suggestions and feedback - nothing major :)

public/app/frontend/src/components/sidebar-block/sidebar-block.html.twig
```diff
@@ -97,22 +97,31 @@ example:
...
-                                options: field.options,
-                                disabled: field.disabled,
-                                error: field.error,
-                            } %}
-                        </div>
+                        {% if field.options %}
+                            <div class="sidebar-block__search-filter-field">
+                                {% include '@components/selection-input/selection-input.html.twig' with {
+                                    type: field.type,
+                                    direction: field.direction,
```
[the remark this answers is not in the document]
EmilyHazlehurst: @EarthlingDavey I didn't know twig allowed the spread operator! Looks like it was added fairly recently. Really good to know, thank you :)

public/app/frontend/src/components/sidebar-block/sidebar-block.html.twig
```diff
@@ -97,22 +97,31 @@ example:
...
         {% elseif variant == 'search-filter' %}
-            <form class="sidebar-block__search-filter">
-                <p class="sidebar_block__search-filter-subtitle">{{ subtitle }}</p>
+            <form
+                class="sidebar-block__search-filter {{ noQuery ? " sidebar-block__search-filter--disabled" : "" }}"
+                {{ noQuery ? " aria-describedby='no-query-hint'" : "" }}>
+                <p class="sidebar_block__search-filter-subtitle" {{ noQuery ? " id='no-query-hint'" : "" }}>
+                    {{ noQuery ? 'Please enter a search query to use filters' : subtitle }}
+                </p>
                 <div class="sidebar-block__search-filter-fields">
```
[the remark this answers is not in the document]
EarthlingDavey: I didn't realise you already used fieldsets, yeah let's not nest them 😁

PR #278, opened 2024-10-14
CDPT-2073: Active indicator is missing on menu links

EmilyHazlehurst:
- Fix logic that displays the indicator (starts with permalink -> permalink starts with)
- Move footer function to the bottom of page/page_home


wilson1000 (approved):
Looks good 😄

PR #288, opened 2024-11-04
Feature/cdpt 1749 - Nav sections

EmilyHazlehurst:


EarthlingDavey:
Hey @EmilyHazlehurst 👋  I think the direction that you're going with this is a good way to do it.

I think there is a big with `renderList`. When I use the editor and write a standard list with text items, and a `pag-nav` list, they are not rendered.

See the screenshot:

image

Some minimal html to replicate this is: 

```html
<!-- wp:list {"className":"is-style-default"} -->
<ul><!-- wp:list-item -->
<li>One</li>
<!-- /wp:list-item --></ul>
<!-- /wp:list -->

<!-- wp:list {"className":"is-style-pag-nav"} -->
<ul><!-- wp:list-item -->
<li>One</li>
<!-- /wp:list-item -->

<!-- wp:list-item -->
<li>Two</li>
<!-- /wp:list-item --></ul>
<!-- /wp:list -->
```

It looks like `$list->getAttribute('class')` is not returning correctly `'is-style-pag-nav'`, so all lists are using the `else { ...` condition.
I don't know why the standard list is not rendering anything, `htmlspecialchars_decode(htmlentities($block['innerHTML']))` is returning some html.

EmilyHazlehurst:
> Hey @EmilyHazlehurst 👋 I think the direction that you're going with this is a good way to do it.
> 
> I think there is a big with `renderList`. When I use the editor and write a standard list with text items, and a `pag-nav` list, they are not rendered.
> 
> See the screenshot:
> 
> image
> Some minimal html to replicate this is:
> 
> ```
> 
> <ul>
> <li>One</li>
> </ul>
> 
> 
> 
> <ul>
> <li>One</li>
> 
> 
> 
> <li>Two</li>
> </ul>
> 
> ```
> 
> It looks like `$list->getAttribute('class')` is not returning `'is-style-pag-nav'` correctly, so all lists are using the `else { ...` condition. I don't know why the standard list is not rendering anything, `htmlspecialchars_decode(htmlentities($block['innerHTML']))` is returning some html.

@EarthlingDavey I was testing this against the existing content on `/courts/procedure-rules/civil/rules`. When I look at that in the backend the block is broken ("Attempt recovery"), but it works with the code I've written on the frontend. 

Interestingly this is true on the live site too, with any new lists the "pag-nav" style isn't applied correctly. Do you know if it's ever worked? It's possible the class was added manually I suppose.

I can't find a consistent answer, but there are a lot of threads about the state of classes on Gutenberg blocks. It looks like classNames aren't supported unless explicitly stated and classNames don't work at all for variants.

Discussion here: https://github.com/WordPress/gutenberg/pull/61864#issuecomment-2233394708

--

I've fixed the issue with the standard list not rendering. I'll make a bug ticket for the classes not being applied properly since it's happening on the live site too.

EarthlingDavey (approved):
Thanks for creating the ticket for the class on the `pag-nav` variant. I'm pretty sure it was a regression that I introduced, by removing className support on all core blocks. 😵‍💫 I've explained more in the ticket, hopefully it should be an easy fix.

The PR looks good - LGTM 🚀

PR #289, opened 2024-11-11
Add overflow-wrap to ensure long links in the body don't stretch the …

EmilyHazlehurst:
Tiny change to stop long links from stretching the page width on mobile.


wilson1000 (approved):
Looks good :)

PR #296, opened 2024-12-11
CDPT- 1855: Update docs

EmilyHazlehurst:
- Bring docs up to date
- Add the beginnings of automated testing with test-runner.js
- Add last reviewed block


public/app/frontend/src/documentation/development/Setup.mdx
```diff
@@ -1,11 +1,27 @@
...
+# Setup
+
+## Frontend only
+Our component library is designed so that it can be worked on without running the backend. If you're working on a
+component and don't need to integrate it with WordPress, then you should only need to:
+
+1. Checkout the justice-gov-uk repository here: https://github.com/ministryofjustice/justice-gov-uk.
+2. `cd` into the `public/app/frontend` directory.
+3. Run `npm install` to install the required packages.
+4. Run `npm run storybook` to start Storybook. It should open automatically in your browser at `localhost:6006`.
```
EarthlingDavey, line 18: As we are running npm run without docker here, should we add a node version to `package.json`?

public/app/frontend/src/documentation/development/Setup.mdx
```diff
@@ -1,11 +1,27 @@
...
+
```
EarthlingDavey, line 19: It's personal view that I try not to run node without using docker. There's a chance that a malicious dependency can run arbitrary scripts during install or runtime.

For that reason I don't have node or npm installed. Could you add this alternative command (or similar) to the docs?

`docker compose run --service-ports node node-frontend`

It will start the node container only, and then run the `node-frontend` task from the makefile. Storybook will then be available at `localhost:6006` or `storybook.justice.docker`.

EarthlingDavey (approved):
Looks good :) Just a couple of questions around npm and a typo.

public/app/frontend/src/documentation/development/Setup.mdx
EmilyHazlehurst: Thanks @EarthlingDavey, that's good to know. I've updated the docs to include that line.

PR #297, opened 2024-12-12
CDPT-2355 Fix errors reported by CLI tool

EarthlingDavey:


deploy/config/local/server.conf
```diff
@@ -138,6 +138,11 @@ server {
         deny all;
     }
 
+    # Deny access to /wp/wp-cron.php /wp/xmlrpc.php, /wp/wp-admin/install.php and /wp/wp-admin/upgrade.php
```
wilson1000, line 141: Hi @EarthlingDavey thanks 😄 
Is this direct access only? 
How does the CRON work on Justice UK?

public/app/themes/justice/style.css
```diff
@@ -4,7 +4,7 @@ Theme URI: https://www.justice.gov.uk/
 Author: Ministry of Justice | Justice Digital | Central Digital Product Team
 Author URI: https://www.gov.uk/government/organisations/ministry-of-justice
 Description: A theme designed to display public-facing, procedural information for the Ministry of Justice, UK
-Version: 1.2
+Version: 10.1.2
```
wilson1000, line 7: Why did we do this?

wilson1000 (approved):
Looks good, some queries but I'm sure you have them covered.

deploy/config/local/server.conf
EarthlingDavey: Thanks for the review :)
This will deny any access for these files via nginx.
Because there aren't any heavy scheduled tasks on Justice UK there wasn't a need to turn off WP_CRON like with the intranet.

public/app/themes/justice/style.css
EarthlingDavey: It doesn't look pretty, but will mean the site looks less interesting during automated scans.

EarthlingDavey: Hi @wilson1000 I'll resolve these 2 conversations so that we can get this merged, but I'm happy to revisit if you think we need to.

PR #298, opened 2024-12-12
CDPT-2368 Fix sentry errors

EarthlingDavey:
This PR fixes recent bugs related to array conversion 

https://ministryofjustice.sentry.io/issues/?environment=production&project=4506813195157504&query=is%3Aresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20array&referrer=issue-list&statsPeriod=7d&utc=true

And, adds a couple of security fixes.

### DIFF
--- a/public/app/themes/justice/inc/search.php
+++ b/public/app/themes/justice/inc/search.php
@@ -60,7 +60,6 @@ class Search
      *
      * @return bool True if the search query is empty, false otherwise.
      */
-
     public function hasEmptyQuery(): bool
     {
         return empty(get_search_query());
@@ -71,7 +70,6 @@ class Search
      *
      * @return int|null The number of search results.
      */
-
     public function getResultCount(): ?int
     {
         if (empty(get_search_query())) {
@@ -89,7 +87,6 @@ class Search
      * @param array $args An array of query parameters to add or modify.
      * @return string The URL for the search results.
      */
-
     public function getSearchUrl($search, $args = [])
     {
         $url_append = '';
@@ -127,7 +124,6 @@ class Search
      *
      * @return array An array of sort options.
      */
-
     public function getSortOptions(): array
     {
         $orderby = get_query_var('orderby');
@@ -151,7 +147,6 @@ class Search
      *
      * @return void
      */
-
     public function redirectOldSearchUrls()
     {
         // Don't redirect if we're in the admin.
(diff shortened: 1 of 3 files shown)


public/app/themes/justice/inc/search.php
```diff
@@ -319,12 +314,77 @@ public function removeColumns(array $columns): array
...
+        }
+    }
+
+
+    /**
+     * Handle malformed search URLs with arrays in the query string.
+     *
+     * This function will redirect the user to the search page if the query string contains arrays.
+     * e.g. /search?audience[$testing]=1 or /search?audience%5B%24testing%5D=1
+     *
```
EmilyHazlehurst, line 364: @EarthlingDavey what's causing this to happen? i.e. what's the flow for a user to end up with an array in the query string or multiple instances of 'page' in the url?

EarthlingDavey: I think it's a vuln. scanner running fuzzing to see what will cause errors.

I don't think we can protect against all fuzzing attacks - some will cause warnings and others will cause errors that will show a generic 500 error page.

But, this one in particular was causing the search plugin to error, and seemed like something that should be avoided. Especially as it has a simple-ish solution that would not get in the way of other site functionality.

EmilyHazlehurst: @EarthlingDavey that makes sense, thanks for clarifying!

EarthlingDavey: No probs :) Thanks for the review

public/app/themes/justice/inc/security.php
```diff
@@ -69,4 +102,48 @@ public function setLoginPeriod(): float|int
...
+    }
+
+    /**
+     * Disable REST API for non-logged in users.
+     *
+     * @see https://developer.wordpress.org/reference/hooks/rest_authentication_errors/
+     *
+     * @param WP_Error|null|true $result
+     * @return WP_Error|null|true
+     */
```
EmilyHazlehurst, line 126: @EarthlingDavey I assume nothing is using the REST API that might break? 
I have a vague memory of there being something reliant on justice.gov.uk post-migration but it might just have been serving styles...

Maybe this one http://xhibit.justice.gov.uk/ ?

EmilyHazlehurst:
One question, but otherwise LGTM 😄

EmilyHazlehurst (approved):

public/app/themes/justice/inc/security.php
EarthlingDavey: Oh yeah, good shout. It was just static assets like js, css & images.

PR #299, opened 2024-12-13
CDPT-2369 Make php-fpm container fs read only.

EarthlingDavey:
This PR makes the filesystem of the php-fpm container readonly. 
It also adds various security contexts to improve security of the pods.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,9 @@ RUN rm zz-docker.conf && \
 ## Set our pool configuration
 COPY deploy/config/php-pool.conf pool.conf
 
+# Create volumes for so that the directories are are writeable when the container is run in read-only mode.
+VOLUME /tmp /var/www/html/public/app/uploads
+
 WORKDIR /var/www/html
 
 ###
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN rm zz-docker.conf && \
 ## Set our pool configuration
 COPY deploy/config/php-pool.conf pool.conf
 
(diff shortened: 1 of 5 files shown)


deploy/staging/deployment.tpl.yml
```diff
@@ -21,22 +21,35 @@ spec:
...
+        seccompProfile:
+          type: "RuntimeDefault"
       containers:
         - name: nginx
           image: ${ECR_URL}:${IMAGE_TAG_NGINX}
           ports:
             - containerPort: 8080
           volumeMounts:
-            - name: uploads
-              mountPath: /var/www/html/public/app/uploads
```
EmilyHazlehurst, line 37: @EarthlingDavey is this okay to remove?

EmilyHazlehurst:
@EarthlingDavey LGTM!

One tiny typo and a question regarding the removal of the uploads directory.

deploy/staging/deployment.tpl.yml
EarthlingDavey: Yes, that's right... it's safe to remove that volume mount because nginx doesn't need access to those files.

Dockerfile
```diff
@@ -22,6 +22,9 @@ RUN rm zz-docker.conf && \
 ## Set our pool configuration
 COPY deploy/config/php-pool.conf pool.conf
 
+# Create volumes for so that the directories are writeable when the container is in read-only mode.
+VOLUME /tmp /var/www/html/public/app/uploads
```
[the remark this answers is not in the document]
EarthlingDavey: TY :)

EmilyHazlehurst (approved):

PR #301, opened 2025-01-13
CDPT-1994: TypeError: e().offset is undefined

EmilyHazlehurst:
- Check whether the banner exists before getting the offset and top position


EarthlingDavey (approved):
Looks good 👍

PR #302, opened 2025-01-13
CDPT-2273: TypeError: l() is not a function

EmilyHazlehurst:
- Remove legacy code referring to a slider component that's no longer in use


EarthlingDavey (approved):
Nice! 💯

PR #303, opened 2025-01-16
CDPT-2268: Horizontal lists no longer have the 'is-style-page-nav' class

EmilyHazlehurst:
- Add an exception to the removeCustomClassNameSupport function for the list block


EarthlingDavey (approved):
Looks good! 💯

PR #304, opened 2025-02-07
Feature/cdpt 2416 justice homepage styling changes

EmilyHazlehurst:

PR #307, opened 2025-03-04
Update Dockerfile

wilson1000:
Snyk error cites an Nginx version mismatch when building the Dockerfile.

This update offers a consistent version provided by a Docker ARG.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+ARG version_nginx=1.26.2
+
 FROM ministryofjustice/wordpress-base-fpm:latest AS base-fpm
 
 # Make the Nginx user available in this container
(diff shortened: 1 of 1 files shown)


EarthlingDavey:
Hey @wilson1000 👋 I tested and built this locally. Your changes look good, but I realised there are 2 bugs and this PR is probably a good opportunity to fix them.

1. As some versions related to gh-pages and jekyll have been updated, there is now an error when building gh-pages. It can be fixed by adding gem "csv" to the end of .github/pages/Gemfile. @EmilyHazlehurst has already done this on the [frontent-rework](https://github.com/ministryofjustice/justice-gov-uk/pull/304#issuecomment-2643316187) branch.
2. I checked locally to see if caching was still working after your change, I discovered that last year I introduced a bug that stopped pages from being cached 😱  The following code block is in `deploy/config/local/server.conf` and `deploy/config/server.conf` and should be deleted.

```
# Skip cache if the concatenated value doesn't contain 200 - the user is not allowed.
# Use concatenation and !~ regex because we cant use the OR operator in nginx
if ($auth_status !~ 200) {
    set $skip_cache 1;
}
```

because the content pages do not use auth, the code block is setting `$skip_cache` for every page. It can be seen by checking the network tab for the justice site, the response headers all say `BYPASS`.

It is safe to delete, as I have found that if a user gets a 401 status code on the login page, the 401 means that the response is not cached by nginx. nginx will only cache 200, 301, & 302.

What do you think about bundling this deletion into your PR?

EarthlingDavey (approved):

PR #315, opened 2025-03-24
CDPT-2574 Temp. disable redis

EarthlingDavey:
This PR disable WP-Redis object cache, because it was increasing P99 latency to 10 seconds.

This PR also adds logging requests that take > ~8s

---

Further investigation is required to work out why the latency is increasing so much for a sub-set of requests.

This should include reviewing slow request logs, and adding logging for cache misses.

### DIFF
--- a/deploy/config/php-pool.conf
+++ b/deploy/config/php-pool.conf
@@ -18,6 +18,10 @@ pm.max_requests = 500
 pm.max_children = 50
 pm.status_path = /status;
 
+; Log a stack trace to stderr for slow queries.
+request_slowlog_timeout = 8s;
+slowlog = /proc/self/fd/2;
+
 [global]
 daemonize = no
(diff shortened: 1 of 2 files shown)


deploy/config/php-pool.conf
```diff
@@ -18,6 +18,10 @@ pm.max_requests = 500
 pm.max_children = 50
 pm.status_path = /status;
 
+; Log a stack trace to stderr for slow queries.
+request_slowlog_timeout = 8s;
+slowlog = /proc/self/fd/2;
+
```
EarthlingDavey, line 24: This is the same as on [intranet's php-pool](https://github.com/ministryofjustice/intranet/blob/main/deploy/config/php-pool.conf)

MeenaModhvadia (approved):
LGTM

PR #327, opened 2025-05-01
CDPT-2416: Help page design improvements

EmilyHazlehurst:
- Use the full grid on homepage, but otherwise limit paragraph width on single sidebar pages


wilson1000 (approved):
Looks good

PR #335, opened 2025-07-18
CDPT-1705 Purge cache after content updates

earthlingdavey:
This PR is a port of the same work done on the Intranet.


deploy/config/local/server.conf
```diff
@@ -137,14 +137,6 @@ server {
         deny all;
     }
 
-    # Deny access to /wp/wp-cron.php, except for connections from the fpm container
-    location ~* ^/wp/wp-cron.php {
-        allow 172.16.0.0/12; # Assuming the fpm container is on this subnet
-        deny all;
-        include /etc/nginx/php-fpm.conf;
-        fastcgi_pass fpm;
-    }
```
wilson1000, line 146: Hey @earthlingdavey - could you expand on why this layer of protection was removed? For example, do we have an alternate method in place?

deploy/development/deployment.tpl.yml
```diff
@@ -137,6 +143,17 @@ spec:
                 secretKeyRef:
                   name: elasticache-output
                   key: auth_token
+            # This is the IP address of the pod, set by Kubernetes.
+            # Pods on the network can use this IP to communicate with each other.
+            # i.e. An nginx purge url can be created with http://<NGINX_IP>:8080/purge/<url>
+            #      Purge urls for all pods, can be called from a single fpm container.
+            # N.B. On Kubernetes, the IP address is the same for all containers in the pod,
+            #      and the IP address assigned to a pod will never change once the pod is running.
+            #      On Docker Compose, the hostname is the name of the service, and used instead of the IP address.
```
wilson1000, line 152: You can also use 'http://localhost:8080/purge/<url>` from inside a container. The network created in a pod respects localhost naming to prevent 'IP sniffing'. Is there a particular issue you've found?

wilson1000 (commented):
Hi 😄 let's catch up

deploy/config/local/server.conf
earthlingdavey: Hi mate, good question ... when it was added, I made the assumption that the local networks would always be on `172.16.0.0/12`.

But it seems like we can't rely on that assumption, since when I was working on this PR, the docker network was using a different subnet.

I figure this protection will be ok to remove for local development, and keep for the Cloud Platform deployments. 

Apologies, I should have mentioned it in the PR notes.

PR #353, opened 2025-08-11
CDPT-2866 Secondary menu label to reflect the open state.

earthlingdavey:


public/app/frontend/src/components/navigation-secondary/index.js
```diff
@@ -12,11 +12,14 @@ export default function () {
...
+        // Toggle the open class on the nav element.
         nav.classList.toggle('navigation-secondary__nav--open');
-        button.setAttribute(
-            'aria-expanded',
-            !(button.getAttribute('aria-expanded') === 'true')
-        );
+        // Set the aria-expanded attribute on the button.
+        button.setAttribute('aria-expanded', willOpen);
+        // Update the text content of the button's visually hidden span.
+        button.querySelector('.visually-hidden').textContent = willOpen ? 'Close secondary ' : 'Open secondary ';
```
wilson1000, line 22: Ok to use jQuery here? 
This is a public site.  `setAttribute()` and `querySelector()` have around 95% support in browsers. We can support 100% simply using jQuery, which is already loaded in the browser and provides cross-browser support.

wilson1000: I've noticed vanilla JS is used throughout.

wilson1000 (approved):

PR #355, opened 2025-08-12
CDPT-2866 Secondary menu label to reflect the open state (amend).

earthlingdavey:
This PR fixes a bug that I introduced in the [previous PR](https://github.com/ministryofjustice/justice-gov-uk/pull/353) - where the value of willOpen is the opposite of what is it intended to be.


wilson1000 (approved):
Good spot!

PR #359, opened 2025-08-19
CDPT-2869 Add svg for logo and hi-res images for logotype

earthlingdavey:


wilson1000 (commented):
Hey @earthlingdavey - have you managed to test this update in Browserstack?

wilson1000:
You can use it for free, I think 🤔

wilson1000:
It's a shame we don't have the same QA/Demo setup on this repo. Maybe I'll get another stack created. Would you support that @earthlingdavey?

earthlingdavey:
Good morning @wilson1000 😀 Thanks for the feedback.

WRT QA/Demo setup, could that decision be postponed until we have had some practical experience with the Intranet QA environment? My thoughts are:

1. There might be some tweaks for the Intranet QA flow whilst we are getting used to it.
2. Cognitive load could get a bit much, this project already has 4 environments, with demo being a detached branch that is demanding manual action to keep in sync with main.
3. When demo is deployed, things will be simpler and creation of a QA environment would be easier to reason about.

---

With the css is the concern 2 fold?

- Using flex - which is potentially unsupported in user's browsers
- Introducing new style rules that might not work as expected?

The thing with `flex` is that, that cat is well and truly out of the hat already 😂 It's been used extensively through the theme. So it maybe the case that a QA/BrowserStack session for the whole of the `frontend-rework` is needed. This could be instead of, or in addition to checking this PR in BrowserStack.

With the concern about introducing new styles, for stuff that's well supported, I've usually test locally in a couple of browsers and resize the window to check responsive styles. Personally I think it's excessive to test this kind of stuff in BrowserStack, I think a quick call about this would be helpful to help share our different perspectives/POVs.

wilson1000:
Hey mate, from what it would have taken to run the site through BrowserStack, you've given me extensive thought and reasons for negating the need. I'll approve.

In addition, I wasn't suggesting you did the work for a QA instance like the Intranet, simply suggesting it might be a good idea in future. Apologies for not explaining that.

earthlingdavey:
Hey @wilson1000 I've tested with BrowserStack and for old versions of Edge, Chroma and FireFox. All good.

The frontend-rework branch does fail for IE11, so let's do a full review at a later date :)

PR #364, opened 2025-08-27
CDPT-2872 side bar navigation amend

earthlingdavey:
In this PR:

**1.** Adds the full URL for the /courts/procedure-rules link, so that it is marked as active and expanded correctly.

Before and after controller update, on the procedure rule page:

image 
image

---

**2.** Correctly styles the active item when it has a sublist

Before and after style update:

image image


wilson1000:
Nice 😄

PR #371, opened 2025-09-01
CDPT-2941 Update `sass`, remove deprecated `@import` and fix duplicated `.icon` blocks in css.

earthlingdavey:
In this PR:

Legacy sass `@import` is replaces with `@use`. `@include` statements are now name-spaced.
Reference: https://sass-lang.com/documentation/breaking-changes/import/

The icon css block has been moved, so that is is processed and transformed to css once - instead of once for each frontend component.

The sass module had been updated, along with webpack config `api: "modern-compiler"`. This fixes a legacy JS api warning.
Reference: https://sass-lang.com/documentation/breaking-changes/legacy-js-api/

### DIFF
--- a/public/app/frontend/base.scss
+++ b/public/app/frontend/base.scss
@@ -1,5 +1,5 @@
 @charset "UTF-8";
-@import "./style.scss";
+@use "style";
 
 html {
     font-size: 16px;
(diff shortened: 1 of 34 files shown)


wilson1000 (approved):
Big update buddy! 😄

PR #374, opened 2025-09-03
CDPT-2947 Fix related pages side-panel doesn't render file download links correctly.

earthlingdavey:
- Update getBreadcrumbs to use label property instead of title (for consistency).
- Move page logic out of the template to a dedicated controller file (for separation of concerns).
- Extract link logic out of `templates.php` into `template-links.php` so functions related to links can be used in the elsewhere (i.e. sidebar).
- Fix breadcrumbs when searching within a parent page.


wilson1000 (approved):
Looks good 😄

PR #375, opened 2025-09-03
CDPT-2941 Add scheduled search re-index

earthlingdavey:
This PR adds a nightly search re-index, so that we don't get into a position where an environment has either: not been indexed or is partially indexed.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
       target: dev
     volumes:
       - theme_dist:/var/www/html/public/app/themes/justice/dist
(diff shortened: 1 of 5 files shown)


wilson1000:
Good morning @earthlingdavey, have you identified an issue with ElasticSearch indexing?

wilson1000:
Or... are we using Relevanssi and it isn't indexing automatically, as it should?

earthlingdavey:
@wilson1000 yeah, it's Relevanssi. I've tagged you in the original issue on Jira.

PR #384, opened 2025-09-16
CDPT-3087 Spelling script amend

earthlingdavey:
@wilson1000 I noticed a minor error in the command.

Discounting that, I've run it on dev and it works well 🎉 Thanks for the reviews :D


wilson1000 (approved):
😄

PR #385, opened 2025-09-16
CDPT-2941 Handle h1 headings un page content.

earthlingdavey:
This PR knocks down the heading sizes if H! is in the user content.

Before: 

image

After:

image

### DIFF
--- a/public/app/frontend/style.scss
+++ b/public/app/frontend/style.scss
@@ -571,6 +571,24 @@ $iconSizes: (
         }
     }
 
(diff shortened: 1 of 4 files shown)


wilson1000:
Cool @earthlingdavey 
Just one thing, the H1 heading at the top is far too big in contrast to the design. Would you happen to know if these have been approved?

wilson1000:
A quick note: Sam Merritt suggested that we can remove the H5 and H6 headings and make the H1-H4 in context size-wise with the site.

PR #486, opened 2026-03-25
Update ingress auth realm (it's invalid with new ingress class)

earthlingdavey:
Fixes:

```
Error from server (BadRequest): error when creating "deploy/development/ingress.yaml": admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: annotation nginx.ingress.kubernetes.io/auth-realm contains invalid value
service/nginx-service unchanged
```


MalcolmVonMoJ (approved):
LGTM

PR #557, opened 2026-07-21
WP and composer package updates

earthlingdavey:


brown-a2 (approved):
i'm guessing 6.9.5 was not affected if you are taking the batch code out

PR #568, opened 2026-08-04
Swap cookie plugins

earthlingdavey:
This PR includes:

- install and styling of the Cookie Compliance plugin
- simplification of `wp-script-localization.php` - the way script localization is done in a way that is compatible with the CSP
- `npm audit fix`
- Fix frontend styling bug
- Delete unused code around serving theme assets from CDN


public/app/themes/justice/inc/wp-script-localization.php
```diff
@@ -8,22 +8,35 @@
...
  */
 class WpScriptLocalization
 {
+    // The handle for the loader script registered in registerLocalizeLoaderScript().
+    const LOADER_HANDLE = 'moj-localize-loader';
+
     // Only allow certain script handles to be modified.
+    // Warning! Keep in sync with mojLocalizedDataEntries in src/js/script-localization.js.
     const ALLOWED_SCRIPT_HANDLES = [
         'ccfw-script',
```
earthlingdavey: Delete

